PR #2, opened 2019-08-29
Added analogy intro & extra questions

kravens:
Thunderbolt analogy to Lightning Network payment.
More beginner question.


GlenCooper (approved):
LGTM

01-what-is.asciidoc
```diff
@@ -2,4 +2,24 @@
...
+As the electric current makes its way through the fractal network of ionized air channels it chooses the path of least resistance to reach its destination, a path is formed between the charged up cloud and the ground of earth below. 
+Then a bright flash of light followed by a loud bang occurs as the current dissipated in the path of ionized air channels creates a hot plasma.
+The Lightning Network is named after this natural phenomenon, where the voltage differences are channel balances and the current is a stream of bitcoin of which ownership is transmitted through those channels.
+When lightning strikes the ground, the potential differences are settled at the cost of dissipating energy and making a loud bang.
+In the Lightning Network this happens when payment channels close, the final state of the channel is transmitted to the Bitcoin network with a fee for the miners to include it in the next block.
+
+
+== Why use the Lightning Network?
+
+A second layer protocol can magnify the amount of transactions possible on the settlement layer, so that the base layer will not be overloaded with transactions as Bitcoin gains more active (transacting) participants.
```
nopara73: > as Bitcoin gains more active (transacting) participants. 

I think it's uneccessarily complex phrasing.

01-what-is.asciidoc
```diff
@@ -2,4 +2,24 @@
...
+The Lightning Network is named after this natural phenomenon, where the voltage differences are channel balances and the current is a stream of bitcoin of which ownership is transmitted through those channels.
+When lightning strikes the ground, the potential differences are settled at the cost of dissipating energy and making a loud bang.
+In the Lightning Network this happens when payment channels close, the final state of the channel is transmitted to the Bitcoin network with a fee for the miners to include it in the next block.
+
+
+== Why use the Lightning Network?
+
+A second layer protocol can magnify the amount of transactions possible on the settlement layer, so that the base layer will not be overloaded with transactions as Bitcoin gains more active (transacting) participants. 
+Instead small value payments can be instantly transmitted in a Lightning Channel, or a chain of channels through the Lightning Network.
+As these channels are not settled on-chain with every transaction, it will become more difficult to differentiate the smaller payments that occured on the channel from eachother.
```
nopara73: typo

```suggestion
As these channels are not settled on-chain with every transaction, it will become more difficult to differentiate the smaller payments that occured on the channel from each other.
```

nopara73: Also I'm not entirely sure it's correct.

01-what-is.asciidoc
```diff
@@ -2,4 +2,24 @@
...
+With this move to a higher protocol layer, payments also gain more privacy and fungibility depending on the chosen route.
+
+== How to use the Lightning Network?
+
+Many interoperable implementations are under active development using the BOLT specifications. 
+It can be run as a Mobile Wallet App on any smartphone, a desktop light-client or routing-node.
```
nopara73: Why the capitalization?

01-what-is.asciidoc
```diff
@@ -2,4 +2,24 @@
...
+A channel of ionized air forms when the voltage or potential difference is greater than the breakdown voltage (..kV) between the different regions in the cloud. 
+As the electric current makes its way through the fractal network of ionized air channels it chooses the path of least resistance to reach its destination, a path is formed between the charged up cloud and the ground of earth below. 
+Then a bright flash of light followed by a loud bang occurs as the current dissipated in the path of ionized air channels creates a hot plasma.
+The Lightning Network is named after this natural phenomenon, where the voltage differences are channel balances and the current is a stream of bitcoin of which ownership is transmitted through those channels.
+When lightning strikes the ground, the potential differences are settled at the cost of dissipating energy and making a loud bang.
+In the Lightning Network this happens when payment channels close, the final state of the channel is transmitted to the Bitcoin network with a fee for the miners to include it in the next block.
```
nopara73, line 52: I learned physics in my native language, therefore I'm having a really difficult time to follow the analogy, due to not being familiar with the terminology. I admit it's a badass analogy, but making readers stop reading from the very first paragraph may not be the best way to go.  
The counter argument count be is that, at this point everyone is paying full attention and whoever chews through this first paragraph will find the following parts easier.  
Anyhow, not against it, but I think it worth a discussion.

01-what-is.asciidoc
kravens: Yeah, I tend to use very technical terms. What would you suggest here?
“As Bitcoin adoption grows in the world” perhaps.

01-what-is.asciidoc
kravens: Lightning Network was capitalized in the title, BOLT is an abreviation. Mobile wallet app should be lowercase indeed!

01-what-is.asciidoc
kravens: I will make a nice visual diagram to go with it, so that the concept is clear even without having learned the physics jargon.

super-e:
Why not starting the "why" with the explanation of the problems the LN tries to solve? For example mentioning problems with scalability, confirmation time, fees cost, etc ?

01-what-is.asciidoc
super-e: Agree with Adam, you risk to lose 90% of readers at the first paragraph :) I have a PhD in physics but still I'm not sure what the point was.

01-what-is.asciidoc
nopara73: ACK

01-what-is.asciidoc
nopara73: Yes, I didn't argue with the LN capitalization.

```suggestion
It can be run as a mobile application, a desktop light-client or routing-node.
```

elkimek:
How about to start with this: https://bitcointalk.org/index.php?topic=2500.msg34211#msg34211 ? :)

01-what-is.asciidoc
bitcoina: I like the analogy, but agree with some of the comments here that its too complex and risks losing readers straight away.   It may be better to start with language such as "lightning is a way to improve upon the bitcoin protocol with additional functionality that increases transaction throughput and reduces settlement time. "

elkimek (approved):
Looks good to me :)

renepickhardt:
> How about to start with this: https://bitcointalk.org/index.php?topic=2500.msg34211#msg34211 ? :)

while I admire Hal for the pioneering work that he did he was obviously wrong on that one. How I understand his post he expected that banks would basically create what we would today call side chains in order to scale bitcoin. I think this is not a good motivation for lightning. 
I would much rather prefer (and will add now) this resource: https://www.metzdowd.com/pipermail/cryptography/2008-November/014814.html which was the first reply. Not sure if random lucky guess or with great knowledge precisely on point.

aantonop:
Too early to be writing the introduction. Typically I write the introduction paragraph LAST, as it needs to reflect the things we learn in each chapter. The first chapter introduction is probably the last thing we write/edit in the book

PR #12, opened 2019-09-03
little correction

boufni95:
the word "all" was written twice


renepickhardt (approved):
Thanks! Don't forget to add yourself to the contributors list. If you don't do that to tomorrow I will merge without it. You could still do it later but it is easiest if it happens with the pull request

renepickhardt (approved):

PR #15, opened 2019-09-04
Update ch01.asciidoc

bitcoina:
Updated first sentence.  

Removed "considered to be", as the fact it is a payment system is not controversial, and changed "the" to "a" as there are multiple payment systems for bitcoin, of which lightning is one..

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -4,7 +4,7 @@
 
 === What is the Lightning Network?
 
-The Lightning Network is considered to be the payment system for Bitcoin.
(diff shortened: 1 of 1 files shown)


renepickhardt (approved):
Definitions are usually a Lorelei bit cautious. Let's see if your suggestion will survive (:

super-e:
Pretty much agree with @bitcoina, was not really sure about that first sentence, seemed that Bitcoin does not have a payment layer and LN plays that role. Like this is more neutral, I would even remove it altogether and go for the definition of the second sentence.

PR #16, opened 2019-09-04
Fix some grammar

awesome-doge:
* Fix uppercase.
* Fix the article.
* Fix commas.
* Fix blanks.
* Fix typos

### DIFF
--- a/glossary.asciidoc
+++ b/glossary.asciidoc
@@ -170,7 +170,7 @@ Gossip Protocol::
     Lightning Network nodes send and receive information about the topology of the Lightning Network through gossip messages which are exchanged with their peers.
     The gossip protocol is mainly defined in BOLT 7 and defines the format of the _node_announcement_, _channel_announcement_ and _channel_update messages_.
     In order to prevent SPAM, node announcement messages will only be forwarded if the node already has a channel and channel announcement messages will only be forwarded if the funding transaction of the channel has been confirmed by the Bitcoin network.
-    Usually Lightning nodes connect with their channel partners, but it is fine connected with any other Lightning node in order to process gossip messages.
+    Usually Lightning nodes connect with their channel partners, but it is fine to connect with any other Lightning node in order to process gossip messages.
 
 hardware wallet::
     A hardware wallet is a special type of bitcoin wallet, which stores the user's private keys in a secure hardware device.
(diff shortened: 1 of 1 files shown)


glossary.asciidoc
```diff
@@ -149,7 +149,7 @@ Ephemeral Key::
     Ephemeral keys are mainly within the SPHINX Mix Format and Onion Routing on the Lightning Network.
     They are generated for each execution of the routing process.
     This increases the security of transported messages or payments.
-    Even if an ephemeral key leaks, only information about a single payment becomes public.
```
bitcoina: "leaks" sounds correct, rather than "leak"?

glossary.asciidoc
```diff
@@ -279,18 +279,18 @@ Payment Channel::
     There are currently three methods known to construct a fully duplex bidirectional payment channel.
     Christian Decker proposed a method in his PhD thesis based on invalidation trees.
     In the Lightning Network whitepaper, Joseph Poon and Tadge Dryja describe the Revocable Sequence Maturity Contract based method that is currently being implemented on the Lightning Network.
-Recently Christian Decker et al. came up with the Eltoo mechanism which would require a Bitcoin softfork.
+Recently Christian Decker et al. Came up with the Eltoo mechanism which would require a Bitcoin softfork.
```
bitcoina: I think lower case "came" is correct, this isn't a new sentence.   maybe it would be better to use the word "proposed" instead of "came up with".  Thx Bitcoina.

awesome-doge: Thanks for the advice.

glossary.asciidoc
```diff
@@ -170,11 +170,11 @@ Gossip Protocol::
     Lightning Network nodes send and receive information about the topology of the Lightning Network through gossip messages which are exchanged with their peers.
     The gossip protocol is mainly defined in BOLT 7 and defines the format of the _node_announcement_, _channel_announcement_ and _channel_update messages_.
     In order to prevent SPAM, node announcement messages will only be forwarded if the node already has a channel and channel announcement messages will only be forwarded if the funding transaction of the channel has been confirmed by the Bitcoin network.
-    Usually Lightning nodes connect with their channel partners, but it is fine connect with any other Lightning node in order to process gossip messages.
+    Usually Lightning nodes connect with their channel partners, but it is fine connected with any other Lightning node in order to process gossip messages.
```
renepickhardt: My initial typo. It was supposed to be "but it is fine to connect...." so the initial wird is correct but there is a word missing

renepickhardt (changes requested):
See comment! Thanks for feedback and review!

renepickhardt (approved):

Volimjasnu12 (commented):
Pozdrav,

Molio bi da mi netko pomogne jer sam u proslosti bio aktivan i sada duzi period nisam.
Sta uciniti?

PR #17, opened 2019-09-04
Update glossary.asciidoc

bjarnemagnussen:
I was just skimming through the gloassary to get a feeling for the topics that may be covered in this book. 

I added small corrections when I found them. Feel free to use them as suggestions, or leave them be. :)


renepickhardt (approved):
It might make sense to also suggest the changes about segwit to mastering bitcoin from where I have taken them over (obviously in a not critical enough way!) thanks for these catches!

PR #20, opened 2019-09-06
Update ch01.asciidoc

bitcoina:
Added info related to proposer and location of initial simple payment channel discussion, useful for expanding out this section.


renepickhardt (approved):
Awesome saved me the time searching for the primary source. I had in mind it was on Bitcoin talk and even earlier. Will merge fürs if there was an earlier version someone will point this out later.

PR #22, opened 2019-09-08
Update ch01.asciidoc

bitcoina:
Added a reference to Mike Hearn as one of the original proposers of uni-directional payment channels.


renepickhardt (changes requested):
just figured out the original link to bitcoin talk which also refered to a blogpost by gavin andresen in july 2012 http://web.archive.org/web/20190730234737/http://gavintech.blogspot.com/web/20150510204329/http://gavintech.blogspot.com/2012/07/off-chain-transactions.html
I will add those resources for later summary.

PR #24, opened 2019-09-09
Update ch01.asciidoc

bitcoina:
Updated the merchant example and put it into the right place.  Thanks @RicardoM17 for your contributions!   I've replaced the names, as I think the Alice / Bob notation brings some baggage with it that we probably don't need.  (Or if we do want to use it, we should re-apply the names in correct order at the end of the writing process.)


Chriton (changes requested):
"costumers" should be "customers"

renepickhardt (changes requested):
I suggest with silke to stick to confirmation time and processing time of payments rather then the fees for her customers. There will be a section about consumers who don't like fees and waiting. but for Silke instant settlement of payments is crucial.

renepickhardt (approved):

PR #25, opened 2019-09-09
Update ch01.asciidoc

bitcoina:
Separated out from previous pull request - (now closed).   Changed the name in this description and incorporated the feedback from @RicardoM17 (thank you!).   Use case is still valid and moving towards being generic enough to include.


renepickhardt (approved):
still very much like this story. I will merge this now and open an issue refering to the real world case so that we can discuss if the name eventually should be changed back to the real person who did this.

PR #26, opened 2019-09-09
Update ch01.asciidoc

bitcoina:
Started to fill out the Getting your First Bitcoin on the Lightning Network.   While custodial solutions are not ideal, they may be a good place for some users to start (given that smaller amounts of money involved) - I think it would be dis-service to the reader not to mention them (and highlight their risks).

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -85,6 +85,16 @@ A merchant will be happy to get an additional payment method which is virtually 
 [[getting_first_bitcoin]]
 ==== Getting Your First Bitcoin on the Lightning Network
 
+Custodial Solutions
+
+Custodial solutions are where a 3rd party has ownership (custody) of your assets, and provides you a service based on this. Custodial solutions are familiar to anyone who has a bank account. In an ideal world, non-custodial lightning networs solutions would be easy enough to use that custodial solutions are not necessary, however in the current state of the crypto-currency lifecycle, there is a place for custodial solutions.  This book will not recommend any specific custodial solutions, however they may be a starting place for people wishing to experiement with the technology. 
+
+Custodial solutions typically run a full-node and lightning implementation in the background, and provide a front end user-interface for customers to send and receive lightning bitcoins.  
+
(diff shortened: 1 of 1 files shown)


ch01.asciidoc
```diff
@@ -85,6 +85,16 @@ A merchant will be happy to get an additional payment method which is virtually
 [[getting_first_bitcoin]]
 ==== Getting Your First Bitcoin on the Lightning Network
 
+Custodial Solutions
+
+Custodial solutions are where a 3rd party has ownership (custody) of your assets, and provides you a service based on this. Custodial solutions are familiar to anyone who has a bank account. In an ideal world, non-custodial lightning networs solutions would be easy enough to use that custodial solutions are not necessary, however in the current state of the crypto-currency lifecycle, there is a place for custodial solutions.  This book will not recommend any specific custodial solutions, however they may be a starting place for people wishing to experiement with the technology.
```
renepickhardt, line 90: "easy enough" is quite judging / subjective. I'd rather have it more objective. 

I like the idea of not recommending a specific custodial solution (:

ch01.asciidoc
```diff
@@ -85,6 +85,16 @@ A merchant will be happy to get an additional payment method which is virtually
 [[getting_first_bitcoin]]
 ==== Getting Your First Bitcoin on the Lightning Network
 
+Custodial Solutions
```
renepickhardt, line 88: as a subheading this should be formatted differently

renepickhardt (commented):
not sure if we should start the book with custodial solutions. I liked the approach of mastering bitcoin where they started with a cellphone bitcoin wallet. Mastering bitcoin also mentioned web wallets (which effectively are custodial solutions)
My guess would be that due to the fast moving innovation taking place in this place we move writing this chapter / section to the very end of the writing process. I will tag this PR as WIP (work in progress)

bitcoina:
Agreed - its a good discussion to have.   Things like BlueWallet are great for getting people into lightning, but also kind of defeat the object. :)

bitcoina:
Agree that we're too early in process to start this  section.   Its no biggie, so closing this PR until we get to it later (keep the PR list clean and relevant.)

PR #27, opened 2019-09-09
Update topics.asciidoc

norisg:
**Adding some topics for the motivation chapter:**

- I thought it would be a good idea to begin the lnbook motivation with the trilemma "security, dezentralization and scaling" => This idea is from a talk of Antonopoulos 

- In addition a comparison to a common fiat second layer system (Visa) might be interesting to draw conclusion for a non tech-savvy person.

### DIFF
--- a/topics.asciidoc
+++ b/topics.asciidoc
@@ -50,7 +50,10 @@ Review of some mechanics and terms from Mastering Bitcoin. This would probably b
 
 === Motivation: Why the Lightning Network
 
-* clear definition what does scaling of a computer system mean? (I realize that discussions often fail because that term is not well defined among participants. A similar definition might be necessary for Decentralized and trustless) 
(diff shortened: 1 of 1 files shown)


renepickhardt (commented):
I would still keep the definition of scaling at the beginning. Also when introducing the trilemma we should also define the terms decentralization and potentially also security. While I am familiar with Andreas talk about this trilemma and it made sense to me I am not sure if it is just an opinion or if that is well described in the scientific literature like the CAP theorem. Maybe we should have include a little bit research about data science, big data, scaling architectures / software systems and see what the definitions and issues are. Obviously blockchain does not fit into the regular scale up / scale out model. 

I already had a quick look at what blogs like highscalability.com offer with respect to blockchain and I was not too happy. It almost seems as if the term was too new.

Would prefer not to merge yet but continue the research in the direction of scalability + blockchain.

PR #30, opened 2019-09-10
Update ch01.asciidoc

bitcoina:
I'm proposing to change the word refugee to economic migrant.   Refugee suggests that someone has had to leave their own country without a choice, where-as I think most people submitting remittances have left their country  by their own choice. 

I think economic migrant is a step towards the right wording, but could also carry other unwanted connotations.  Maybe "remittance sender" would be more accurate and less controversial...


renepickhardt:
What about just using the term `migrant` without economical?

bitcoina:
agree!

renepickhardt (approved):

PR #31, opened 2019-09-10
Update ch01.asciidoc

bitcoina:
Proposed some edits to the initial paragraphs on the 'nSequence' number.  Aim is to make it slightly more accurate and readable.   Happy to break these proposals down into smaller chunks if there is too much in a single pull request.


renepickhardt (approved):
Will merge I am not too happy with it yet as I kind of disagree with the intro. I am not sure if satoshi recognized this to be a second layer protocol. or if he just thought it would be good if a transaction could have several versions before it confirms to allow for a more real / time like system.

bitcoina:
Agree - hopefully it moved it on a little bit!   I think I read somewhere that satoshi thought it would allow high-frequency trading, so I think he was aware of the second layer aspect to it.    I'll do some digging.

PR #32, opened 2019-09-10
Update ch01.asciidoc

bitcoina:
Couple of edits to the great Ultra Transaction server piece. - Didn't know about this proposal before.   Gavin Andresen was a true innovator.


renepickhardt (approved):
ack since only stylistic changes

PR #41, opened 2019-09-11
Fix intent

nopara73:


renepickhardt (changes requested):
there was a typo that you caught. but your suggestion might be wrong. A could have an interest to publish old state and "get the money back". The word `and` was meant to connect both sentences. So I guess splitting into two sentences would be the way to go

nopara73:
Like this?

renepickhardt (changes requested):
the entire mess started with me choosing A as the subject instead of B which now with your change became even clearer to me. I will fix this myself and merge now. Thanks for helping to find a semantic problem! Those are usually the hard ones as the grammar usually does not raise a flag

renepickhardt:
@nopara73 I rewrote basically the entire paragraph to make it more clear. I hope there should be no errors. but fill free to add another review and reopen a new PR if you encounter a problem.

PR #43, opened 2019-09-11
`all Bitcoin` -> `all bitcoins`

nopara73:
I'm pretty sure I've read the correct pluralization and capitalization in the Mastering Bitcoin book a long time ago.

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -40,7 +40,7 @@ Effectively the most recent update of the payment channel (encoded as a bitcoin 
 In July of 2011, on the bitcointalk.org forum,  a pseudonomous user by the name of _hashcoin_ proposed the usage of Timelocks via the `nLockTime` function of the bitcoin network to solve the custody problem of exchanges.footnote:[Hashcoin on Bitcoin talk on July 4th 2011 - Instant TX for established business relationships (need replacements/nLockTime) http://web.archive.org/web/20190419103503/https://bitcointalk.org/index.php?topic=25786.0]
 The idea was to be able to quickly trade / sell Bitcoin without the necessity to send all the bitcoins to the exchange in the first place.
 Hashcoin effectively proposed what we would call an unidirectional payment channel today.
-With this mechanism a user A could fund a multisig wallet between A and another user B together with a timelocked transaction sending all Bitcoin Back to A.
+With this mechanism a user A could fund a multisig wallet between A and another user B together with a timelocked transaction sending all bitcoins Back to A.
 The funding tx would not be signed and broadcasted by A before B provided a signature for the spend.
 Hashcoin imagined user B to be an exchange.
 Whenever user A wanted to send Bitcoin to user B user A could create a newer spend of the funding transaction which would send less Bitcoin back to A and more bitcoin to B.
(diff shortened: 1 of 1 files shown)


renepickhardt (commented):
when do we write bitcoin with a capital B and when with a lowercase b? is it as soon was we speek of the concret monetary unit of exchange / the token the letter b will be lower case whereas if we refer to the name of the protocol / currency the latter will be capitalized?

ugamkamat:
@renepickhardt yes, Bitcoin with capital B is generally used to represent technology, while bitcoins with small b is used to represent currency. There is an interesting discussion on bitcoin meta stackexchange on this [here](https://bitcoin.meta.stackexchange.com/questions/21/how-do-we-capitalize-bitcoin-or-bitcoin-or-bitcoin).

renepickhardt:
Then there would still remain the question for plural s. Is it 5 bitcoin or 5 bitcoins?

ugamkamat:
When referring as a currency (bitcoins), I think the convention should be the same as dollars.

wombat26:
_Bitcoin_ (with capital B) is normally uses to refer to the whole protocol, technology, network, etc.

_bitcoin_ ( with lower b) is normally uses to name the digital token unit, the coin, the currency. On the economic / trading jargon people often says 0.5 bitcoin, 1 bitcoin, 2 bitcoin, 10 bitcoin.

In _"Mastering Bitcoin"_  is used _bitcoin_ with lower b and singular to refer to any amount of bitcoin. Even if a prefix is used the singular mode remains e.g. 100 millibitcoin 

On the other hand, the word _satoshi_ is used in singular and plural,
1 satoshi, 5 satoshis, 100 satoshis.

_My name is Anibal Santaella, a pleasure to be here._

wombat26:
> Then there would still remain the question for plural s. Is it 5 bitcoin or 5 bitcoins?
In _"Mastering Bitcoin"_  is used bitcoin with lower b and singular to refer to any amount of bitcoin.  
Even if a prefix is used the singular mode remains e.g. 100 millibitcoin.

nopara73:
Maybe always plural, except when at 1 BTC?

- 0.1 bitcoins
- 1 bitcoin
- 1.1 bitcoins
- 21,000,000 bitcoins

renepickhardt:
other currencies don't have a plural when being added as a unit. 

I say this costs 5 Euro or 5 Dollar. I don't say 5 Euros or 5 Dollars. would only use the plural if I have several coins / bills in my pocket but I am not even sure if that was colloquial. In bitcoin this situation would not really have an equivalent unless I would refer to several UTXOs which I guess noone ever did when utilizing the plural s or not (as the UTXOs are hardly visible to the user)

Still hesitant to merge because there still seems to be a disagreement about plural.

wombat26:
I think we have to keep on mind the economy, simplicity, syntax of the language while stablishing these definitions.

ugamkamat:
@renepickhardt I disagree. We say "the coffee cost me $2 dollars" rather than "the coffee cost me 2 dollar". The use of plural depends crucially on whether the phrase is prenominal (placed before a noun) or not. I tried to search something on a sister stackexchange site to back my reasoning. Found one [here](https://ell.stackexchange.com/questions/104829/correct-way-to-express-monetary-loss-dollar-vs-dollars)

nopara73:
> Still hesitant to merge because there still seems to be a disagreement about plural.

I'm not insisting on my pluralization, so I'll wait for the decision.

PatrickZGW:
Using currencies without a plural is the standard way in other languages like German. In English the plural sounds more natural to me.

Chriton:
I found this pdf here https://www.oecd-ilibrary.org/economics/oecd-style-guide_9789264243439-en. At pages 70-73 they have examples of different currencyes and their plural. Usually currencies have a plural form which is different from the singular.
Also another article is here http://www.amigaforever.com/kb/13-144. See the **Singular vs. Plural Form** section.

aantonop:
I prefer the singular form, as bitcoin is intangible and based on arbitrary sized UTXO values. This make me think it should be treated as an uncountable noun, rather than a noun describing discreet units. 

I used this convention (mostly but not 100% consistently) in Mastering Bitcoin and would like to continue to use it here:

1 bitcoin
2 bitcoin
many bitcoin
She sent all of her bitcoin to Bob
There will only ever be less than 21 million bitcoin
I have 3.2 bitcoin

etc.

PR #47, opened 2019-09-11
Replace "wallet" with "address", in relation to hashcash's unilateral channel proposal.

bitcoina:
Propose to replace the comment regarding multi-sig wallet, with multi-sig transaction.


renepickhardt (changes requested):
the transaction is not funded. so I reject this. instead wallet we could say address. I agree that wallet is not the most specific tearm

bitcoina:
Learning stuff here.... :) - its fun..

i.e.  how a multisig address is created and how you fund it.  Thanks,

renepickhardt (approved):
will be hard for me in future. as I usually use the term multisignature wallet but I guess in this case multisignature address is just more accurate (: 

it is by the way hashcoins unilateral channel proposal :P

bitcoina:
Glad you are keeping us all honest :)

PR #48, opened 2019-09-11
Added few extra topics

ugamkamat:
@renepickhardt  Nice work on the outline. Here are the changes that I have done. Would appreciate your feedback. I have added a couple of topics like `nLocktime`, `CLTV`, `CSV`, and `nSequence` to the "Review of Bitcoin" section as I feel they form the foundation of the transactions in Lightning Network. Also, I have added the topic "Need for a payment receipt" in the "Motivation of Lightning Network" section as invoice payments gives us a cryptographic proof of payment that is generally lacking in a normal on-chain transaction. 

I changed the name of `Breach Remedy Transaction` to `Penalty transaction` as the BRT term does not appear anywhere outside Joseph Poon's whitepaper. I think you should also rethink the RSMC term as that term also does not appear in the BOLTs and outside the whitepaper.

I think it would be worthwhile to add scriptless scripts and benefits Taproot can provide to Lightning Network in BOLT 1.1 /2.0 section. With Taproot, we would be able to do wonderful things with Lightning onion and as the book gets published in 2H20, we would be might be much closer to Taproot implementation on mainchain.

### DIFF
--- a/topics.asciidoc
+++ b/topics.asciidoc
@@ -27,6 +27,7 @@ Review of some mechanics and terms from Mastering Bitcoin. This would probably b
 * fees
 * hash
 * Hash Function
+* Locktime and CLTV
 * multisignature
 * output
 * P2PKH
@@ -36,6 +37,7 @@ Review of some mechanics and terms from Mastering Bitcoin. This would probably b
 * P2WSH
 * Proof-of-Work
 * RIPEMD-160
+* Sequence and CSV
 * Script
 * ScriptPubKey (aka pubkey script)
 * ScriptSig (aka signature script)
@@ -54,6 +56,7 @@ Review of some mechanics and terms from Mastering Bitcoin. This would probably b
 * Scale Payments (amount of possible payments independent of block size / data base size)
 * increase privacy (not all payments stored on chain)
 * decrease settlement time (get rid of confirmations)
+* need of payment receipt (normally lacking in a Bitcoin mainchain transaction)
 
 === Using / transacting Bitcoin with the Lightning Network
 This group of topics would be end user oriented and is rather non technical
@@ -106,7 +109,7 @@ While also important for end users to be aware of some of the topics here it is 
 A summary of how the Lightning Network protocol works.
 
 * BOLT
-* Breach Remedy Transaction
+* Penalty Transaction
 * Closing Transaction
 * Commitment Transaction
 * Contract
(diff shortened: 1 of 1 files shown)


topics.asciidoc
```diff
@@ -54,6 +56,7 @@ Review of some mechanics and terms from Mastering Bitcoin. This would probably b
 * Scale Payments (amount of possible payments independent of block size / data base size)
 * increase privacy (not all payments stored on chain)
 * decrease settlement time (get rid of confirmations)
+* need of payment receipt (normally lacking in a Bitcoin mainchain transaction)
```
renepickhardt: Do you know the routing section of my slide deck? I introduce the invoice based system to explain how it is currently needed for the routing layer. While a getting started section similar to Mastering Bitcoin will certainly explain the payment process I am not sure if we need it in the motivation chapter. In general I agree that bolt 11 and invoices / receipts will need special care

renepickhardt (commented):
Great additions will most likely merge as we are still in the brainstorming phase. Just not sure about the location of need for receipts.

topics.asciidoc
ugamkamat: @renepickhardt Yeah, I was torn when including the receipt point in that section. I now agree with your reasoning that payment receipt was not the *motivation* behind Lightning Network, but more of an added benefit that it provides. You can probably explain the importance of receipt if you include a section on spontaneous payments.

ugamkamat: I pushed a commit removing the invoice point in motivation, so that it is easier for you to directly merge. Also added channel factories point in BOLT 1.1/2.0 section

ugamkamat: Also added channel factories in Bolt 1.1/2.0 section

topics.asciidoc
```diff
@@ -139,12 +141,14 @@ A summary of how the Lightning Network protocol works.
 === BOLT 1.1 / 2.0
 Future trends and developments on the Lightning Network.
 
+* channel factories
```
renepickhardt: Was already there as part of eltoo. It is pretty clear that they went fine with the penalty based asymmetrical protocol as the communication overhead would be too severe

renepickhardt (changes requested):
Nit: remove double entry please

topics.asciidoc
ugamkamat: @renepickhardt  that was my bad. Probably was too late in the night for me when I made the changes :). I removed it.

renepickhardt (approved):
thanks! as mentioned before excellent contributions / suggestions.

ugamkamat:
My pleasure!

PR #50, opened 2019-09-12
added dandelion under privacy to topics list

bitcoina:
Not sure if this has another name, if so, please reject.   I just added dandelion under privacy in the topics list.

### DIFF
--- a/topics.asciidoc
+++ b/topics.asciidoc
@@ -162,6 +162,7 @@ Everything about the Lightning Network that is not covered by the BOLTs but impo
 * graph (theory)
 * lnd
 * privacy
+** dandelion
 * social network analysis
 * Topology
 * Traffic analysis


renepickhardt:
Could you elaborate how this relates to the lightning network?

Reading https://medium.com/@thecryptoconomy/dandelions-and-a-bright-future-for-bitcoin-privacy-712dbc4b1ec5 I don't see how this would actually impact privacy on lightning. Lightnings gossip protocol signs messages with noise public key anyway and the privacy model is pretty different to the one in Bitcoin

bitcoina:
Ah, I thought that dandelion was incorporated into lightning network protocol.  Apologies for time wasting.  closing pull request.

nopara73:
Both Dandelion and Sphinx (which is an onion routing protocol) came from the original Chaum Mixing network research, but they took different direction, so I don't think it's justified to add.

PR #51, opened 2019-09-12
Stylistic change to Meni Rosenfeld paragraph.

bitcoina:
Propose to change the word order and a few words to make it easier to read.


renepickhardt (changes requested):
think the commas are much better grammar but the `and` should also be removed in the proposition. will do so now and merge

Issue #36, opened 2019-09-11
Use of LN as Abbreviation


nopara73:
It'll be written million times in the book. Maybe it would make sense to consider to use `LN` consistently instead of `Lightning Network` after the first paragraph.  

Maybe it would be even better done so to use `Lightning Network` in titles and `LN` in bodies.

aantonop:
Agreed. O'Reilly style guide is to associate the acronym to a definition once, then use the acronym. 

LN everywhere after the first paragraph.

PR #56, opened 2019-09-16
Use LN abbreviation for Lightning Network (glossary file)

yahiheb:
As suggested in #36


renepickhardt:
I am not sure about the use of that abbreviation in general. But I think in the glossary we should not do it as the O'Reilly style is that the first time a word appears it is written out and then abbreviated for the remainder of that section However in a glossary people usually don't reed from alpha through omega but rather look up single terms. Thus I would rather reject this PR. any thoughts @aantonop

aantonop:
Agreed. Let's not do it in the glossary or appendixes

PR #57, opened 2019-09-16
Use LN abbreviation for Lightning Network (ch01 file)

yahiheb:
As suggested in #36


ch01.asciidoc
```diff
@@ -6,18 +6,18 @@
 
 The Lightning Network is a payment system for Bitcoin.
```
nopara73:
```suggestion
The Lightning Network (LN) is a payment system for Bitcoin.
```

nopara73: The abbreviation should be introduced at the first usage (dismiss my commit suggestion if it was already introduced previously and I missed it.)

nopara73 (approved):

renepickhardt (approved):
still not too happy about using this abbreviation throughout the book but I guess it relly will increase the readability.

PR #61, opened 2019-09-16
travis: Add linter to detect "Satoshi style" double spacing

practicalswift:
Add linter to detect "Satoshi style" double spacing.


renepickhardt (approved):
I learnt an important lesson today from you. Thanks!

PR #75, opened 2019-10-01
One small language fix

Moller40:
Change-Id: I2be0befb5063834d3b27ff22e585647ab22d236d


ch01.asciidoc
```diff
@@ -110,7 +110,7 @@ Scalable Off-Chain Instant Payments https://lightning.network/lightning-network-
 
 As an electronic cash system it preserves the 3 most important properties of money (medium of exchange, store of value, and unit of account).
 The invention of money (and in particular Bitcoin) was primarily made to facilitate trade and enable the exchange of value between people.
-However, without the LN, Bitcoin is hard to be used concurrently by millions of people.
```
Chriton, line 113: @Moller40 I think the text was ok initially, the new phrase doesn't seems to be right

Chriton (commented):
I think the text was ok initially, the new phrase doesn't seems to be right

PR #85, opened 2019-11-05
Update ch01.asciidoc

elima-iii:
line 58:

originally: "This mechanism would allow two users to engage into several smaller transactions..."

corrected: "This mechanism would allow to users to engage in several smaller transactions..."


matthiasdebernardini:
I think you meant  "This mechanism would allow ~~to~~ users to engage in several smaller transactions..." in the correction.

PR #88, opened 2019-11-08
AMA reorg edits to ch01 

aantonop:
Chapters 1 & 2 will be less technical, to help non-technical users get a high level view of LN.  To make more accessible, I am removing technical detail, reducing the history segment, and introducing "motivation" and "basics" sections.

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -119,7 +119,7 @@ Scalable Off-Chain Instant Payments https://lightning.network/lightning-network-
 [[user-stories]]
 === Lightning Network Uses, Users, and Their Stories
 
-As an electronic cash system it preserves the 3 most important properties of money (medium of exchange, store of value, and unit of account).
+As an electronic cash system Bitcoin preserves the 3 most important properties of money (medium of exchange, store of value, and unit of account).
 The invention of money (and in particular Bitcoin) was primarily made to facilitate trade and enable the exchange of value between people.
 However, without the LN, Bitcoin is hard to be used concurrently by millions of people.
 Therefore, in order to fully understand the uses of the LN, we'll examine it from the perspective of people using it.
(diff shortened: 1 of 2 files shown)


aantonop:
In progress: still pushing commits to this

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+== Getting Started
+
+// TODO intro paragraph - no empty headings!
+
+==== Choosing a Lightning Network Wallet
+
+* full nodes (c-lightning, eclair, lnd) + remote controls
+* phone / desktop wallets (SPV clients)
+* custodial services / wallets?
+// Mastering bitcoin also had a section about custodial web wallets. So it might be fair to include them.
```
renepickhardt, line 11: yes but I kind of didn't like that section about mastering bitcoin (:

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
+[[ch02_getting_started]]
+== Getting Started
+
+// TODO intro paragraph - no empty headings!
+
+==== Choosing a Lightning Network Wallet
+
+* full nodes (c-lightning, eclair, lnd) + remote controls
```
renepickhardt, line 8: does eclair have a remote control? I am not aware of one.

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+* phone / desktop wallets (SPV clients)
+* custodial services / wallets?
+// Mastering bitcoin also had a section about custodial web wallets. So it might be fair to include them.
+
+==== Quick Start
+
+* We start with a mobile wallet like eclair
```
renepickhardt, line 15: BTW it does not and will never run on iOS because of the JVM

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+* Verify the authenticity of the software (e.g. App store / Acinq website / github)
+** discuss that a general phising scheme might consist of tricking you to download a similar looking software
+
+[[getting_first_bitcoin]]
+==== Getting Your First Bitcoin on the Lightning Network
+
+* Trade fiat for Bitcoin (as in Mastering Bitcoin)
```
renepickhardt, line 22: Will this be onchain to eclair and then open the channel yourself or will the other person open the channel with you and push liquidity? I would suggest the second option (as we can't go for turbo channels yet)

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+
+[[using_own_bitcoin]]
+==== Process for people who already own Bitcoin ====
+
+* send bitcoin to lightning wallet (1 onchain transaction - soon nodes / wallets may support funding a channel directly without sending bitcoins to the lightning network wallet first)
+* find a node to open a channel with (Node explorer / Autopilots / ...)
```
renepickhardt, line 28: gossip protocol? the node explorer is nothing but a visualization of `lightning-cli listnodes` we could also talk about dns seeds but I guess that is way too technical for here

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+* open a connection
```
renepickhardt, line 29: needs to be done in the onramp case too (even if the other side opens the channel one of the two has to open a connection first)

I remmeber that I found this extremely confusing why I needed to open a connection first when I didn't understand the details of the protocol

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+* open a channel
+* wait confirmations for the channel to become operational
+
+[[sending_receiving]]
+==== Sending and Receiving Bitcoin on the Lightning Network
+* have the person who helps onbording open a payment channel (As complex as a bitcoin transaction)
+* showcase how the funds now can be send back and forth
```
renepickhardt, line 36: mentioning the channel reserve? also here when I did my first experiment I was wondering why I could not the 1 satoshi back that I had received

renepickhardt (commented):
just some thoughts about the intended structure of chapter 02. I guess you have already all of this in your mind. but I thought I could just drop it while I was looking at it.

ch01.asciidoc
```diff
@@ -4,15 +4,20 @@
 
 === What is the Lightning Network?
```
renepickhardt, line 5: chapter 3 also has an intro chapter that explains lightning again on a slightly more technical level (c.f.: https://github.com/lnbook/lnbook/blob/develop/ch03.asciidoc ) is the redundancy an issue? I personally don't think so. However we should align the statements. For example you removed the "payment system" part which I forgot when writing chapter 3. Yet I think that by the end of the day this is THE crucial property and main reason why wo do all of that lightning stuff anyway.

ch02.asciidoc
Roasbeef: Remote control as in an RPC interface? It _does_ have a JSON-RPC interface which can be used to drive a node by a local program. Technically c-lightning is the same as well since their primary interface is accessible only via a unix socket, though I think they intend for user-defined plugs-in to fill that gap.

ch02.asciidoc
Roasbeef: I'm guessing just but Bitcoin elsewhere, and send it to your wallet of choice.

ch02.asciidoc
Roasbeef: There really aren't any "web" wallets for Lightning as is today, but there're custodial wallets that are mobile-only (BlueWallet, Wallet of Satoshi, etc).

ch02.asciidoc
Roasbeef: If we want to cover both platforms, then there're few wallets on iOS that we can choose from.

ch02.asciidoc
Roasbeef: I think maybe we can just reference it here, but then go dive deeper into the ramifications of channel reserves as a whole.

ch02.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+** create invoice
+** decode invoice (do always before paying!)
+** pay an invoice
+* showcase how an arbitrary invoice can be paid e.g. https://voting.ln.rene-pickhardt.de
```
Roasbeef, line 40: How is this different than a regular invoice? Do you mean a zero-valued invoice @renepickhardt ?

renepickhardt: I think a regular invoice was meant

aantonop:
I'm going to merge this and proceed with the rest of my chapter 1 edits using direct merges to develop instead of pull requests

PR #89, opened 2019-11-08
typo, clearer phrasing

boufni95:


renepickhardt (commented):
NACK. This list is currently for bulletpoints anyway and the text is about to be written. so I don't see the point of fixing the grammar here.

PR #94, opened 2019-12-11
ch01-03: rolling PR of WIP edits 

Roasbeef:
Just wanted to throw this up so we can start some discussion. I'll be updating it continually as I progress through the referenced chapters. I've tried to keep the commits localized to a sentence or logical change to make the PR easier to follow, and so we can easily remove commits without modifying the greater PR.

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -177,11 +177,13 @@ In this section we will start by choosing the right software to demonstrate the 
 
 The Lightning Network is accessed via software applications that can speak the Lightning Network protocol. A _Lightning Network Node_ (or simply "node") is a software application that communicates on a peer-to-peer basis with other LN nodes, forming the Lightning _Network_ itself. Nodes also include "wallet" functionality, so they can send and receive payments over the LN and on the Bitcoin network.
 
-Lightning network nodes also need access to the Bitcoin blockchain, or another blockchain such as Litecoin. Users have the highest degree of control by running their own authoritative Bitcoin node and LN node. However, LN nodes can also rely on Bitcoin nodes operated by a third party, by using a lightweight protocol such as _Simplified Payment Verification (SPV)_ or the _Neutrino_ protocol.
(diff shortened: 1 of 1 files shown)


ch01.asciidoc
```diff
@@ -191,4 +206,4 @@ Here are some current examples of LN node and wallet applications for different
 | lnd | Server | Full Node | Optional | Full Control |
 | c-lightning | Server | Full Node | Full Node | Full Control |
 | Zap Desktop | Desktop | Full Node | Full Node | Full Control |
-| Eclair | Mobile | Lightweight |
+| Eclair | Mobile | Lightweight | 3rd Party | Full Control |
```
Roasbeef, line 209: So in this table, I think there's a bit more nuance we can capture here, by using the following columns (traits of each wallet):
  * node implementation 
  * device (mobile vs desktop)
  * blockchain data (full node vs SPV)
  * full channel graph storage (full or partial) 
  * local path finding (in device or delegated) 
  * key management (custodial vs non-custodial)
  * channel creation (manual vs autopilot vs hub/LSP)

ch01.asciidoc
```diff
@@ -62,7 +71,11 @@ Therefore, Bob can securely receive more updates to the channel balance as long
 
 This mechanism would allow two users to engage in several smaller transactions which all happened outside of the Bitcoin network.
 While this construction of the unidirectional payment channel would have solved the custody problem of exchanges it has never been widely implemented.
-We can only speculate for reasons and guess that the overhead communication would have had to be standardized - as it is nowadays in the LN specification - which might have been too much overhead in the early days of Bitcoin.
+
+A fundamental blocker to deploying payment channels in production was the issue of transaction malleability.
```
Roasbeef, line 75: I'm thinking in the series above (attempt at secure channels by construction), we can introduce malleability as the final barrier after being able to enforce commitment revocation.

aantonop:
Merging the in-progress changes, so we can move to drafting Chapters 1 and 2

ch01.asciidoc
aantonop: What do you mean by "node implementation" here? I like this structure a lot and want to explore it in this chapter

PR #96, opened 2019-12-30
Add eclair to the list of server implementations

t-bast:
Even though Eclair is known for its mobile wallets, its primary work
is a server implementation aimed at scalable enterprise deployments.
It doesn't seem fair to not even include it in the list.

I also slightly changed what's in the "Bitcoin Node" column.
Full Node vs Optional didn't make a lot of sense to me (all LN nodes
really need some kind of bitcoin full node).
I don't know if specifying the exact node implementation (like I did)
makes sense, let me know if you have other ideas.

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -187,8 +187,9 @@ Lightning wallets can be installed on a variety of devices, including laptops, s
 Here are some current examples of LN node and wallet applications for different types of devices:
 
 |===
-| Application | Device | LN Node | Bitcoin Node | Wallet Type |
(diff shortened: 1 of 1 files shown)


t-bast:
Note that it may be confusing to have both `eclair` and `eclair-mobile` in the list (especially since it's not meant to be exhaustive). Since the wallet world is evolving very rapidly, I'd be in favor of keeping the line about `eclair` as a server implementation (I don't expect many new server implementations popping up - I do expect many new wallets to appear).

aantonop:
Thank you @t-bast

PR #98, opened 2020-01-10
Upper/lower case for the "B" in "Bitcoin" unified

CandleHater:
Written as bitcoin with a "b" in lowercase usually refers to the currency unit. A capitalized first letter addresses the network or protocol.

I changed every mentioning of "Bitcoin" in this document accordingly and extended the Bitcoin section to include this distinction. If the change is accepted, I'd check the other files regarding this.

### DIFF
--- a/glossary.asciidoc
+++ b/glossary.asciidoc
@@ -1,10 +1,10 @@
 [preface]
 == Quick Glossary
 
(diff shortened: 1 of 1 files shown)


aantonop:
Thanks @CandleHater

PR #103, opened 2020-01-22
CH1 minor edits and suggestions

DariusParvin:
At times the Lightning Network is framed as blockchain agnostic, but then most of the time it's assumed we're talking about a Bitcoin LN node. For consistency and brevity, I feel like it's best to mention at the start that the LN is blockchain agnostic, but from then on assume we're using Bitcoin.

There was some inconsistency for the capitalization of "Lightning Wallet". After it's introduction, I went with lowercase, but I think "LN wallet" would also be an option, consistent with "LN node".

I'm not sure specifically what an "Authoritative" Bitcoin node is, but I feel like it would be implied with just "Bitcoin node" and would be less confusing for the reader.

The oxford comma, used before the "and" at the end of a list, is standard in american english.

This is my first PR, so any feedback is welcome!

### DIFF
--- a/ch01.asciidoc
+++ b/ch01.asciidoc
@@ -4,13 +4,13 @@
 
 === What is the Lightning Network?
 
-The Lightning Network (LN) is a "second layer" protocol on top of Bitcoin and other blockchains that enables fast, secure, private, trustless and permissionless payments.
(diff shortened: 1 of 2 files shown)


aantonop:
Great corrections, thanks!

PR #109, opened 2020-01-23
Change "bitcoins" and "Bitcoins" to "bitcoin"

### LINKED ISSUE #104: bitcoin vs bitcoins in plural form
Both "bitcoin" and "bitcoins" are used in plural form throughout the book. We should decide on a plural form to use going forward. Andreas uses "bitcoin" in the plural in Mastering Bitcoin while Jimmy Song uses "bitcoins" in Programming Bitcoin. Either form is OK as long as it's used consistently.

bolatovumar:
fix #104


aantonop:
Thanks @bolatovumar

PR #150, opened 2020-02-16
add Transaction Malleability

0xmountaintop:


glossary.asciidoc
```diff
@@ -447,7 +447,8 @@ Transaction::
     Transactions can also contain contracts and should not be confused with payments.
 
 Transaction Malleability::
-    TBD.
+    Transaction Malleability is a property that a transaction's hash can change without real changes in transaction content. For example, an unsigned transaction's hash will change after signing, which is inconvenient and may lead to issues.
```
renepickhardt: NIT: hash of a transaction...
Formulation: ... without real changes... maybe use without changing the semantic of the transaction? 
Is that really the best example? What about altering the signature which would be the case that is prevented in the channel opening protocol?

renepickhardt (changes requested):
see the comment not too happy about the example and I think the formulation is unlucky. I am not fully convinced by my suggestion to use `semantic` either.

glossary.asciidoc
0xmountaintop: Hi @renepickhardt  I have make some changes. will be great if you can take a look!

renepickhardt (approved):
still not too happy with the language but I will merge this for now and create an issue that someone might refactor the language in the future. The information is correct (:

PR #151, opened 2020-02-16
fix broken footnote link in 02_getting_started

theborakompanioni:
- replace "https://acinq.io" with "https://acinq.co" (last check on 2020-02-16)


renepickhardt:
ACK

PR #154, opened 2020-02-17
add "Ligtning Messages" section

0xmountaintop:
I have to admit the "Lightning Messages" section I add in this PR still seems insubstantial. Would like to hear from @renepickhardt for your suggestions on skeleton

### DIFF
--- a/communications_protocol.asciidoc
+++ b/communications_protocol.asciidoc
@@ -32,7 +32,7 @@ Lightning Messages follows two forethoughtful rules: "it's OK to be odd" rule an
 
 === "It's OK to be odd" rule
 
-Lightning Message type follows the _it's ok to be odd_ rule. A node is allowed to send a message of an _odd_-numbered type, but not _even_-numbered type without prior negotiation. Once a node receives a _oddly_-but-unknown-typed message, it should ignore it. However, if it receives a _evenly_-but-unknown-typed message, it is expected to fail the channels.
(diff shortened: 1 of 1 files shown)


0xmountaintop:
Not sure whether I should add "it's ok to be odd". will that be too detailed?

renepickhardt (commented):
While I certainly want to talk about Lightning messages and also about the "it's OK to be odd" rule I think this is too early / technical in the sense that chapter 3 is supposed to give a rough overfiew. A lot of the stuff that I have written so far seems rather too technical at this time and for this chapter. I suggest to move the paragraph to a new file maybe named lightning messages so that we can reuse it later when it fits the planned narrative.

0xmountaintop:
> While I certainly want to talk about Lightning messages and also about the "it's OK to be odd" rule I think this is too early / technical in the sense that chapter 3 is supposed to give a rough overfiew. A lot of the stuff that I have written so far seems rather too technical at this time and for this chapter. I suggest to move the paragraph to a new file maybe named lightning messages so that we can reuse it later when it fits the planned narrative.

Yep I definitively agree. And FYI currently it's in communications_protocol chapter, instead of chap03. :P

0xmountaintop:
Hi @renepickhardt , that's my idea for the "Ligtning Messages" section, can you take a look again please? Thanks!

renepickhardt (commented):
The inormation is correct but I pull @Roasbeef into the discussion for the sake of how deep do we want to go into Lightning messages

aantonop:
Thanks @HAOYUatHZ, sorry for the delay!

PR #156, opened 2020-02-20
Fix typos ch03

PatrickZGW:
A couple of things I'd like your opinion on:

- Should it be "Bitcoin Network" or "Bitcoin network"? In my opinion "Bitcoin network" is correct, because the network is not part of the official name. Otherwise we would also have to write "Bitcoin Transaction", "Bitcoin Code", etc.
- What is the correct capitalization of "Onion"? I didn't change those because I wasn't sure what version is correct. I just think it should ever be  consistently capitalized or not.


bolatovumar:
@PatrickLemke it should be `Bitcoin network` and `onion routing` IMO. See usage in Wikipedia for example: https://en.wikipedia.org/wiki/Onion_routing

aantonop:
@PatrickLemke If you can fix the conflicts, ready to merge this!

PatrickZGW:
I think the Travis CI build is still failing because of an issue in another file? Should this be fixed in this PR or can this be merged as is?

aantonop:
Merge as is as soon as deconflicted

PatrickZGW:
Merge conflicts should be resolved.

PR #164, opened 2020-02-24
WIP: Development Environment: Docker and docker-compose setup

### LINKED ISSUE #158: Development environment standardization
In order to make it easier for users to follow the examples, we have made two decisions to standardize the development environment. These decisions will be implemented as part of Chapter 4

1) We will be using a Docker container to build a consistent environment in which all the examples can be run and tested. The Dockerfile will be shared on this repository and instructions to run it will be included in the text

2) We will be using Bitcoin in REGTEST mode, and providing instructions on how to deploy and use regtest and LN on top of it. This will alleviate the need to bootstrap a node and help developers understand the power of using regtest in a dev environment. 

Maybe one day we can "backport" both of these ideas into Mastering Bitcoin, if we execute successfully here...

Topics for discussion:

* Is Docker the right choice?

 I think it will help users of Windows, Mac and various other systems load a standardized dev environment. Though many developers in our space run some version of Linux, there are also many readers who do not and would find it difficult to replicate our examples. We can't offer tools and examples for every OS, so we have to pick one

* If Docker, what OS and image

I think we should standardize on Ubuntu 18.04 LTS, because it is the most supported OS across all the open source projects. As an LTS version (Long Term Support), it will be supported until April 2023. By then we will be in a second edition and can move to the next LTS version. The downside is that an Ubuntu 18.04 image is much "heavier" than say alpine, but the purpose of this container is to be very very standard and have the most well known tools, so it is worth paying the price in disk space for Ubuntu. 

* How much of the dev environment should be packaged in the Docker image?

I think at least the basic dev tools and pre-requisite libraries for the various tools we will install (eg. all the things you apt install before bitcoin, Go support for lnd, etc). Do we also pre-install bitcoind latest from PPA? Or do we include that in the examples so that people can learn how to do it on their own? Mastering Bitcoin already covers all that in chapter 3 and I don't want to repeat too much. 

* Separate repository?

Should we make a separate repository for the Docker image, scripts and examples, so that we can clone it and make it part of the Docker container, without having to pull in the book itself? I think so....

Please add your thoughts and suggestions!

aantonop:
Initial attempt at Dockerfile for standardized dev environment

See #158

Installs dev tools, bitcoind, c-lightning, lnd


renepickhardt:
what about Rust Lightning (becoming LDK by square crypto), Electrum and https://github.com/nayutaco/ptarmigan ?

bolatovumar:
@renepickhardt isn't Rust Lightning still not really a full implementation of the LN protocol and way behind LND/c-lightning/Eclair? It probably will be more feature-complete by the time this book is published though.

code/docker/bitcoind-base/Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM ubuntu:18.04
```
fguisso, line 1: Golang have some good official images than ubuntu https://hub.docker.com/_/golang
With Debian, windowsservercore or alpine, you don't need to install the binaries and configs

St333p:
What about [lighter](https://gitlab.com/inbitcoin/lighter)?

It would allow users to have a uniform interface across implementation (c-lightning, eclair, lnd supported; electrum coming soon and ptarmigan in roadmap) at least for the most common operations. Of course one still needs to directly talk to c-lightning to use plugins, but that is gonna be 5% of readers. 

I might also find some time to write about it and to fix the _compose_ we are using for internal tests for publication if there is interest in this regard.

code/docker/bitcoind-base/Dockerfile
0xmountaintop: agree. and this can reduce the docker image size.

code/docker/c-lightning-base/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:18.04
```
0xmountaintop, line 1: we use bitcoind's prebuilt docker image but we build lightningd image by ourselves. will it be a good idea to also use   c-lightning/lnd's docker image?

lnd's: https://github.com/lightningnetwork/lnd/blob/master/docker/README.md

it also provide a good tutorial on how nodes interact with each other, and [how to run the image](https://github.com/lightningnetwork/lnd/blob/master/docs/DOCKER.md)

0xmountaintop: I just search.   seems  there's no official  docker images of   c-lightning /  lnd    published on docker hub.  I believe that's the reason @aantonop choose to build it by ourselves.

alternatives approaches can be:

+ use https://hub.docker.com/r/cdecker/lightning-ci/dockerfile
+ or maybe @Roasbeef  can publish a docker image onto docker hub on behalf of lightninglabs?

code/docker/devenv/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+# Go 1.13 for LND
+RUN curl -SL https://dl.google.com/go/go1.13.linux-amd64.tar.gz \
+	| tar -xzC /usr/local
+ENV PATH $PATH:/usr/local/go/bin
+ENV GOPATH /gocode
+RUN mkdir -p $GOPATH
+ENV PATH $PATH:$GOPATH/bin
+
+# LND
+RUN go get -d github.com/lightningnetwork/lnd
```
0xmountaintop, line 54: will it be good if we specify a version (using commit hash or ,,,).   

My concern is that,    there may be some breaking changes, and the commands may become incompatible. for example, https://en.bitcoin.it/wiki/Original_Bitcoin_client/API_calls_list these are deprecated.

PR #165, opened 2020-02-25
Add simple build instructions for the book to README.md

sradforth:
Never see asciidocs format before, added to the README so others know how to use/edit this correctly.
Not sure if you want others to build it locally or not but if you do I think this could help get more people contributing as I was stuck on where to start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,27 @@ The book is suitable for technical readers with an understanding of the fundamen
 
 The authors welcome contributions to this book! Read the [Guide to Contributing](CONTRIBUTING.md)
 
+## Building
+
+This book is written using the open source asciidoctor mark up language.
+To build the book locally you can install asciidoctor package following the steps found at
+
+    https://asciidoctor.org/
+
+
+On a command line you can type
+
+```    asciidoctor *.asciidoc -D output```
+
+and copy the image files in with
+
+```    cp images/ output/```
+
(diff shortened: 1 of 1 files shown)


aantonop:
The license does not allow building (derivatives) until one year after publication, when we hope to open it up to CC BY SA.

Until then, building rendered output as PDF is discouraged as it violates the publishers rights and makes them less likely to promote open source as they do now.

renepickhardt:
@aantonop the information is however useful. so how to we proceed here?

aantonop:
Until the license allows, I will not add instructions that help people
violate the license and cause the publisher to avoid future open source
projects.

When the license is suitable, I will add a template and Makefile to build
PDF, HTML, ePUB

sradforth:
I fully understand and agree with that but perhaps put some instructions so it's obvious how people can read/edit/review... I was new to this so wasn't sure how to start contributing.

Is there a simple suggested way to review/edit it as that will probably be really helpful to share.

aantonop:
> I fully understand and agree with that but perhaps put some instructions so it's obvious how people can read/edit/review... I was new to this so wasn't sure how to start contributing.
> 
> Is there a simple suggested way to review/edit it as that will probably be really helpful to share.

Definitely. We will be building out the contributor guide and the readme really soon!

Thanks @sradforth

PR #169, opened 2020-04-01
Update preface.asciidoc

odomyk:


renepickhardt (changes requested):
could not find any other commits / PRs under this account. thus I reject the PR. please correct me if I oversaw something

PR #170, opened 2020-04-15
minor improvements and clarifications

8go:
- minor clarifications
- incorrect use of plural
- repetitive use of "this way" --> this fashion
- clarifying "each time" --> each payment
- added detail that LN txs are not stored permanently
- repetitive use of "explanation" --> clarification
- seeing --> revisiting
- etc.


renepickhardt (approved):
thanks a lot

PR #173, opened 2020-04-15
clarifications and additions

8go:
some modifications to improve, 

some clarifications

some additions to add more details where appropriate

... lightweight Bitcoin client (commonly referred to as Simplified Payment Verification (SPV)) to partially validate the correctness of their blockchain. --> partially? Probably you mean "indirectly" as you delegate to a third party. Also SPVs clients don't have a blockchain, so why "their blockchain"? --> to reduce this confusion --> lightweight Bitcoin client (commonly referred to as Simplified Payment Verification (SPV)) to interact with the blockchain

... lightning wallets ... --> ... Lightning wallets ...

centralized and added more details to definition of non-custodial wallet

etc


02_getting_started.asciidoc
```diff
@@ -16,38 +16,38 @@ The term "Lightning Wallet" is somewhat ambiguous, as it can describe a broad va
...
 
-Every other component of a lightning wallet brings similar considerations of trust. If all the components are under the control of the user, then the amount of trust in third parties is minimized, bringing maximum power to the user. Of course, this is a direct trade-off, as with that power comes the responsibility to manage complex software.
+Every other component of a Lightning wallet brings similar considerations of trust. If all the components are under the control of the user, then the amount of trust in third parties is minimized, bringing maximum power to the user. Of course, this brings a direct trade-off, as with that power comes the corresponding responsibility to manage complex software.
 
-Every user must consider their own technical skills before deciding what type of lightning wallet to use. Those with strong technical skills should use a lightning wallet that puts all of the components under the direct control of the user. Those with less technical skill but a desire to control their funds, should choose a _non-custodial_ lightning wallet, even if some of the components (other than the keystore) rely on some trusted third parties.
```
renepickhardt, line 33: I would not agree to remove this relative sentence

02_getting_started.asciidoc
```diff
@@ -70,29 +70,29 @@ In <<lnwallet-examples>> we see some examples of currently popular Lightning nod
...
+Alice notices a few different elements on this page, that help her ascertain that this is, most likely, the correct "Eclair Mobile" wallet she is looking for. Firstly, the organization "ACINQ" footnote:[ACINQ: Developers of the Eclair Mobile Lightning wallet (https://acinq.co/).] is listed as the developer of this mobile wallet, which Alice knows from her research is the correct developer. Secondly, the wallet has been installed "10,000+" times and has more than 320 positive reviews. It is unlikely this is a rogue app that has snuck into the Play Store. As third step, she goes to the ACINQ website (https://acinq.co/). There she goes to the Download section or looks for the link to the Google App store. She finds the link and clicks it. She compares that this link brings her to the very same app in the Google App Store. Satisfied by these findings, Alice installs the Eclair app on her mobile device.
```
renepickhardt, line 95: I like the idea of verifying via the ACINQ website. however this website could also be just in that moment compromized... security is really tricky. but that is more a general concern

renepickhardt (changes requested):
95% ack

02_getting_started.asciidoc
8go: Of course, the website could be hacked, re-routed or faked. Still the check is worth while. 

Best check is to verify hash and GPG signature of .APK package but a) many products do not provide this info and b) many new users are not capable of doing this check correctly. In contrary, checking the website is trivial and improves security a little bit.

PR #177, opened 2020-04-16
improved definition of LN

8go:
- removed incorrect comma in front of "as well as"
- added "both" to underline that LN has 2 main functionalities
- "... on the network." leaves it undefined. The reader might ask to which network it refers to? To the before mentioned network of payment-channels? Added "Bitcoin" to clarify.


renepickhardt (approved):
very good! I assume we will refine this definition even more

PR #196, opened 2020-04-21
minor addition to chapter "Examining the ugly way"

8go:
- an addendum explaining in more detail how to catch a cheat
- explain the responsibilities and the options

### DIFF
--- a/ch03.asciidoc
+++ b/ch03.asciidoc
@@ -211,6 +211,7 @@ The timelock is usually measured in blocktime and can be up to 2016.
 On average, 2016 blocks represent or last two weeks (assuming a blocktime of 10 minutes as in the Bitcoin network).
 Be aware, the exact number of blocks used for the timeout can be set to different values by different software.
 2016 is just the maximum value, but smaller values are possible.
+footnote:[Users likely prefer a longer timeout value for a larger channel capacity.]
 Within the timelock anyone who knows the revocation secret can spend the output even before the timelock is over.
 Alice and Bob know only one half each of the revocation secret.
 But if they share their half with the other channel partner, then the other partner knows the full secret.
--- a/ch03.asciidoc
+++ b/ch03.asciidoc
@@ -200,6 +200,8 @@ Usually the commitment transaction has at least two outputs, one for each partne
 However a channel partner will encumber their own output with a timelock and a revocation secret.
 The timelock prevents the owner of the output to spend it directly once the commitment transaction was included in a block.
 The timelock is usually measured in blocktime and can be up to 2016 which is statistically speaking two weeks (assuming a blocktime of 10 minutes which is the target for the Bitcoin network).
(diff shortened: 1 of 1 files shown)


ch03.asciidoc
```diff
@@ -286,6 +286,10 @@ In that case the fees of the commitment transaction are again paid by the partne
 Also all the routing attempts will have to be resolved just as during the force close.
 While this method could be fully executed faster than the good and the bad way to close the channel, it is obviously never recommended to engage in this channel closing protocol.
 
+So, how do you catch a cheat or a protocol breach in your day-to-day activities? You do so by running software that monitors the public Bitcoin blockchain for on-line transactions that correspond to any of your commitment transaction. A Lightning node can do this monitoring for you. Specific Lightning wallets might do this for you. Running a Lightning node 24x7 will catch all cheats. But even if you run your Lightning node only once every Sunday, it will still catch all cheats because the timelock is roughly two weeks. In addition, you can delegate this responsibility to someone else. Someone else can monitor the Bitcoin blockchain for you. One of these technologies is called a "Watchtower".
```
renepickhardt: we should emphasze that the two weeks are only true if the lightning network was configured in this way and channels have been set up with long enough time locks. in general this statement does not need to be true

renepickhardt (commented):
we need to be more careful when geving security advices

ch03.asciidoc
8go: OK, agreed. I added more text to make the risks more clear and give enough warning.
The new changes are under Commit # 2 under this PR 
commit 232a81a

8go:
Commit # 2 in 232a81a did the requested changes.
Commit # 3 resolves conflicts. https://github.com/lnbook/lnbook/pull/196/commits/b03cecdd7bccd6164f7c1daa203511e9e7cf8a82 

Should be ready to pull. Kind regards.

ch03.asciidoc
```diff
@@ -209,9 +209,11 @@ As new addition, a channel partner will encumber their own output with a timeloc
 The timelock prevents the owner of the output to spend it immediately once the commitment transaction was included in a block.
 The timelock is usually measured in blocktime and can be up to 2016.
 On average, 2016 blocks represent or last two weeks (assuming a blocktime of 10 minutes as in the Bitcoin network).
+Be aware, the exact number of blocks used for the timeout can be set to different values by different software.
+2016 is just the maximum value, but smaller values are possible.
```
Roasbeef, line 213: Can mention here that one likely wants a longer value the larger the channel size. 2016 is just what's used as the canonical example.

8go: suggestion incorporated

renepickhardt (approved):
Those are nice additions. We might move part of the text to a different chapter as it is very detailed for an overview but I will merge for now

PR #197, opened 2020-04-21
minor improvements to chapter "Invoices"

8go:
- changed the order a bit.
- since "invoice" is introduced it should be explained right away
- addition, explaining differences in concepts to Bitcoin


ch03.asciidoc
```diff
@@ -289,17 +289,21 @@ While this method could be fully executed faster than the good and the bad way t
 === Invoices
 
 Every payment on the Lightning Network starts with a person who wants to receive bitcoins issuing an invoice.
+An invoice is nothing more than a very simple payment instruction containing information such as a recipient and an amount.
```
renepickhardt: and the payment hash. actually the payment hash ist the most important pieace of information (:

ch03.asciidoc
```diff
@@ -289,17 +289,21 @@ While this method could be fully executed faster than the good and the bad way t
...
+The payee has to pass the invoice to the payer via any arbitrary means.
+This could be done via a Webservice, an instant message on a chat app, an SMS or an Email.
+Invoices are usually encoded either as long bech32-encoded strings or as QR codes which enables them to be easily scanned by smartphones.
+The invoices contains the amount of bitcoin that is requested and a signature of the payee.
 The later is used to extract the address of the payee so that the payer knows whom to send the money.
 Besides some other meta data the most important but not quite obvious data in the invoice is a Payment Hash.
 
+Did you notice how this contrasts with Bitcoin and how different language terms are used? In Bitcoin, the recipient passes an address to the sender. In Lightning, the payee creates an invoice and sends an invoice to the payer. In Bitcoin, the payer sends funds to an address. In Lightning, the payer pays an invoice and it gets routed to the payee. Bitcoin is based on the concept of an "address", and Lightning is a payment network and based on the concept of an "invoice".
```
renepickhardt, line 356: while we are at it we could also stress that on bitcoin we have transactions while on lightning we have payments

renepickhardt (commented):
two ideas for further improvements. other than that LGTM

ch03.asciidoc
8go: I wanted to keep it simple, but the hash is so crucial that it should be included. Good point.
see commit # 2 97b20db5e032f8fc56c243b2ee2c4818c8e1ec49

ch03.asciidoc
8go: I thought about it, but put it into the "Differences to Bitcoin" section. 
I was afraid it would get repetitive.
Still, it is just 1 sentence more. 
I added it. 
See commit # 3 in 9a737b111b96cd202f645b8e1a190d75e40ffd82

8go:
All done. 
See Commit 2 and 3 in 97b20db5e032f8fc56c243b2ee2c4818c8e1ec49 and 9a737b111b96cd202f645b8e1a190d75e40ffd82

Ready for review and pull. Kind regards.

ch03.asciidoc
```diff
@@ -289,17 +289,22 @@ While this method could be fully executed faster than the good and the bad way t
 === Invoices
 
 Every payment on the Lightning Network starts with a person who wants to receive bitcoins issuing an invoice.
+An invoice is nothing more than a very simple payment instruction containing information such as a unique payment identifier, called a payment hash, a recipient and an amount.
```
Roasbeef: Also contains other extra details which may be needed in order to complete the payment. One example if the `payment_addr` field for MPP payments.

8go: adjusted text to make it more general and open-ended.

renepickhardt (approved):
ACK

PR #201, opened 2020-04-21
minor improvements to chapter "Onion Routing"

8go:
- improvement of English
- some additions
- etc.

### DIFF
--- a/ch03.asciidoc
+++ b/ch03.asciidoc
@@ -386,18 +386,25 @@ This includes which public payment channels are connecting nodes, which capacity
 ==== Onion routing
 
 If the sending node of a payment has selected a path that is supposed to be used to make the payment, the Lightning Network uses an onion routing scheme similar to the famous TOR-network.
-The routing scheme is called the SPHINX mixformat and will be explained in detail in a later chapter.
-For now we want to focus on its properties for the transport of payments which we also call onions.
+The routing scheme is called the "SPHINX mixformat" and will be explained in detail in a later chapter.
+For now we want to focus on its properties for the transport of payments.
+A payment package used for routing is called an "onion".
+It is called an "onion" due to the historic use of this word related to the Tor network which is also called the Onion network.
+Imagine the payment package as an onion. On its route from payer to payee it is handed from node to node. 
+As will be explained, each node peals off the outermost layer of this onion and then forwards the remaining, but one layer smaller, onion to the next node.
 
-1. The most important property is that a routing node can only see on which channels it received an onion and on which channel to set up an HTLCs and thus to which peer to forward the onion. This means that no routing node can know who initiated the payment and for whom the payment is supposed to be. The exception of course would be if the node is the recipient. In that case it would know that it was the final destination.
-2. The onions are small enough to fit into a single TCP/IP package and actually even a link layer frame. This will make traffic analysis for intruding the privacy of the payments almost impossible.
-3. The onions are constructed in a way that they will always have the same length independent of the position of the processing node along the path.
-4. Onions can have up to 20 hops included allowing for sufficiently long paths.
-5. The encryption of the onion for every hop uses different ephemeral encryption keys with every single onion. Should a key (in particular the private key of the public node key) leak at some point in time an attacker who collected onions cannot decrypt the other onions that have been stored.
+1. The most important property is that a routing node can only see on which channel it received an onion and on which channel to set up an HTLCs and thus to which peer to forward the onion. This means that no routing node can know who initiated the payment and to whom the payment is destined. The exception is the payee. The payee knows that it is the final destination. 
+This achieves privacy.
+2. The onions are small enough to fit into a single TCP/IP package and actually even a link layer frame. This will make traffic analysis of the payments almost impossible, increasing privacy further. 
+3. The onions are constructed in a way that they will always have the same length independent of the position of the processing node along the path. This is also done to maintain privacy.
+4. Onions can have up to 20 hops or onion layers if you prefer. This allows for sufficiently long paths.
+5. The encryption of the onion for every hop uses different ephemeral encryption keys with every single onion. Should a key (in particular the private key of the public node key) leak at some point in time an attacker who collected onions cannot decrypt other onions.
+In simpler terms, keys are never reused in order to achieve more security.
 6. Errors can be sent back from the erring node in an encrypted way to the original sender. This is particularly useful as we have seen that Lightning nodes who initiate the onions select a path without knowing whether every node has enough liquidity along their channels to forward the payment.
 
-As mentioned we will discuss the details of the Onion Format later but we note already that the `Payment Hash`, while needed to set up the HTLCs for the payment, is not transported within the onions.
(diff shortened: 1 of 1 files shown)


ch03.asciidoc
```diff
@@ -386,18 +386,25 @@ This includes which public payment channels are connecting nodes, which capacity
...
+In simpler terms, keys are never reused in order to achieve more security.
 6. Errors can be sent back from the erring node in an encrypted way to the original sender. This is particularly useful as we have seen that Lightning nodes who initiate the onions select a path without knowing whether every node has enough liquidity along their channels to forward the payment.
 
-As mentioned we will discuss the details of the Onion Format later but we note already that the `Payment Hash`, while needed to set up the HTLCs for the payment, is not transported within the onions.
-The `Payment Hash` is rather included in the Lightning Message that also transports the onion.
+We will discuss the details of the Onion Format later. But we note already that the `Payment Hash` which is required to set up the HTLCs for the payment is not transported within the onions.
+The `Payment Hash` is rather included in the Lightning Message alongside the onion.
+This is done to make the Payment Hash convient to obtain without the need to decrypt any onion.
```
renepickhardt: I would not agree that this is the reason. I think it is rather about the size of the onion and not have to include redundant data

ch03.asciidoc
```diff
@@ -386,18 +386,25 @@ This includes which public payment channels are connecting nodes, which capacity
...
-2. The onions are small enough to fit into a single TCP/IP package and actually even a link layer frame. This will make traffic analysis for intruding the privacy of the payments almost impossible.
-3. The onions are constructed in a way that they will always have the same length independent of the position of the processing node along the path.
-4. Onions can have up to 20 hops included allowing for sufficiently long paths.
-5. The encryption of the onion for every hop uses different ephemeral encryption keys with every single onion. Should a key (in particular the private key of the public node key) leak at some point in time an attacker who collected onions cannot decrypt the other onions that have been stored.
+The routing scheme is called the "SPHINX mixformat" and will be explained in detail in a later chapter.
+For now we want to focus on its properties for the transport of payments.
+A payment package used for routing is called an "onion".
+It is called an "onion" due to the historic use of this word related to the Tor network which is also called the Onion network.
```
renepickhardt: you explain afterwards how messages are pealed and that they are called onions for that reason. so I am not sure if the historic reason is how we should present this.

ch03.asciidoc
```diff
@@ -386,18 +386,25 @@ This includes which public payment channels are connecting nodes, which capacity
...
+Imagine the payment package as an onion. On its route from payer to payee it is handed from node to node. 
+As will be explained, each node peals off the outermost layer of this onion and then forwards the remaining, but one layer smaller, onion to the next node.
```
renepickhardt: the onions are all of same size which is a particular feature of the SPHINX mix format. so I feel not comfortable about this statement in this form

renepickhardt (commented):
could you please address the 3 comments I made? in particular the second one about the fixed length of the onions

ch03.asciidoc
8go: It is done most likely because of both reasons: faster access, smaller overall size.
Since I cannot quantify reduction in size, I left that part out. 
Just stated the obvious and correct fact, without interpreting it.

See commit # 1 in 75cfd8d1b265c0255e25bbd19a21aa08f02b99ca

ch03.asciidoc
8go: I do believe that the term comes from Tor and has been used there extensively for 17 years. 
I rephrased because this is more of a side note, it is not crucial to the discussion of LN.
See commit # 3 in a968296e8051c3dfff9f4a9b30403a171d9611c5

ch03.asciidoc
8go: I thought exactly the same as I wrote it. That it could lead to a wrong misinterpretation. 
OK. I clarified it further.
See commit # 4 in 45c09fc25ffd890b5e5b07d28bc164d937a6c654

8go:
All done. 3 more commits:
75cfd8d1b265c0255e25bbd19a21aa08f02b99ca
a968296e8051c3dfff9f4a9b30403a171d9611c5
45c09fc25ffd890b5e5b07d28bc164d937a6c654

Hope you agree. Kind regards.

ch03.asciidoc
```diff
@@ -386,18 +386,42 @@ This includes which public payment channels are connecting nodes, which capacity
...
+The box is used in this analogy to express that the size of the "package" sent along the route is always the same. 
+The first recipient receives a box of the same size as the last recipient, the payee.
+This is done to create privacy. 
+If not done, an observer could analyze the size of the communication and deduce how many hops there are on the route. 
+If the size does not change there is no way to tell if this is the first or the last hop in the route.
+The box in our analogy is what technically is called an "onion" in the Lightning network.
+
+1. The most important property is that a routing node can only see on which channel it received an onion and on which channel to set up an HTLCs and thus to which peer to forward the onion. This means that no routing node can know who initiated the payment and to whom the payment is destined. The exception is the payee. The payee knows that it is the final destination. 
+This achieves privacy.
+2. The onions are small enough to fit into a single TCP/IP package and actually even a link layer frame. This will make traffic analysis of the payments almost impossible, increasing privacy further.
```
Roasbeef: I know this was left over, but this is a pretty strong claim that we need to modify later on. Onion routing alone isn't enough to thwart traffic analysis.

8go: I adjusted the expectations by modifying 3 sentences in 3 bullet points.

ch03.asciidoc
```diff
@@ -386,18 +386,47 @@ This includes which public payment channels are connecting nodes, which capacity
...
+For now we want to focus on its properties for the transport of payments.
+A payment package used for routing is called an "onion".
+footnote:[The term "onion" was originally used by the TOR project.
+The acronym TOR stands for "The Onion Router".
+Moreover, the Tor network is also called the Onion network. 
+The Tor project uses an onion as its logo and the top level domain name used by TOR services on the internet is ".onion".]
+As a simplified physical analogy imagine the payment package as a sealed box with a onion inside. 
+On its route from payer to payee the sealed box is handed from node to node. 
+Sealed just means that the box is encrypted and only the receiving node can open it.
+As will be explained, each node opens the box (using a cryptographic key), takes out the onion, and peals off the outermost layer of this onion.
```
renepickhardt: actually the opening and taking out of the onion payload does not require a cryptographic key. it is rather the pealing off that requires this

ch03.asciidoc
```diff
@@ -386,18 +386,47 @@ This includes which public payment channels are connecting nodes, which capacity
...
+The receiving node can only read this outermost layer, no further layers, achieved through cryptography.
+This outermost layer tells the node if the box needs to be routed further and to which node.
```
renepickhardt: along which channel (which implies the node)

renepickhardt (changes requested):
if I can't change this on github I will merge and make a subsequent change as a different commit as this is only very small details

PR #204, opened 2020-04-22
minor improvements to chapter "Comparison with Bitcoin" part 1

8go:
- no commas when subject is the same on both parts of sentence
- nothing seems to be gained by introducing concept of "traffic" --> removed it
- instead explained what the scarce resources are
- more explanations
- etc.


Roasbeef (approved):
LGTM 💎

PR #208, opened 2020-04-22
addition to chapter "Comparison with Bitcoin" part 5

8go:
- additional chapter

### DIFF
--- a/ch03.asciidoc
+++ b/ch03.asciidoc
@@ -469,6 +469,23 @@ Since Bob can directly close the channel, which costs fees paid by Alice, she wi
 
 While the Lightning Network is built on top of Bitcoin, and inherits many of its features and properties, there are important differences that users of both need to be aware of.
 
+==== Addresses vs Invoices, Transactions vs Payments
+
+Bitcoin and Lightning have different terminology.
+In Bitcoin, a typical operation is that a user receives a Bitcoin address (e.g. he reads it on a webpage, or he received it in an instant message from a friend).
+He then uses his Bitcoin wallet to create a transaction to send funds to this address. 
+
+In the Lightning Network all starts at the payee. 
+The payee creates an invoice. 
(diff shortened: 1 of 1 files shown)


ch03.asciidoc
```diff
@@ -469,6 +469,23 @@ Since Bob can directly close the channel, which costs fees paid by Alice, she wi
...
+
+In the Lightning Network all starts at the payee. 
+The payee creates an invoice. 
+A Lightning "invoice" could be seen as an analogy to a Bitcoin "address".
+The payee provides the Lightning invoice to the payer. 
+This is similar to Bitcoin, the payee places it on the webpage at checkout, or the payee is a friend and sends it via instant message or email. 
+The payer uses his Lightning wallet to make the invoice payment (e.g. by scanning a QR code). 
+In Lightning, "payment" is the analog term for a Bitcoin "transaction".
+
+Along the same lines, in Lightning we prefer to use "payer" and "payee" to what Bitcoin often refers to as "sender" and "receiver".
```
Roasbeef: Would say everything looks good other than this sentence. It seems pretty arbitrary and just a matter of convention. I myself typically use sender/receiver when talking about LN, either at the code/spec level or in abstract.

Roasbeef (changes requested):

ch03.asciidoc
8go: You are certainly right, it is a mere conventions. I made some changes to soften this part. Still, as a convention it helps people thinking about LN as a payment network. 

Also, with the optional direct-address send within an invoice, we have an option that is more sender-receiver like. But I did not want to complicate things further. in the non-technical chapter. Details will come later at the appropriate place. 

Improved in Commit # 2 in 12a9200c05e94b220c7c52076fa7571a923824dc

renepickhardt (approved):
like the softened approach though I am not necessarily a fan of pushing suggestions for conventions that do not seem to be consent yet in such an anvironment.

PR #209, opened 2020-04-22
addition to chapter "Comparison with Bitcoin" part 6

8go:
- added new chapter


Roasbeef (approved):
Nice addition, not sure this belongs at this point in the chapter, but we're planning to re-organize things anyway.

Issue #176, opened 2020-04-15
Editorial:  Re-write Categories of LN Wallets in chapter 2


8go:
Chapter 2 has this categorization: 

[[lnwallet-categories]]
.Broad Categories of "LN Wallets"
|===
| Wallet Type          | LN Node        | Keystore/Custody | Technical Skill
| Full Node & Wallet   | Full Node      | Non-Custodial | High
| Non-Custodial Wallet | 3rd-party node | Non-Custodial | Medium
| Custodial Wallet     | 3rd-party node | Custodial     | Low
|===

It is not perfect. It is not ideal. It sort of implies that a "Full Node & Wallet" is not a "Non-Custodial Wallet" because they are 2 different categories. It also states that a "Non-Custodial Wallet" uses "3rd-party node" for LN Node which is not the true definition. In order to simplify, things are getting mixed up here and we are _not_ communicating a clear message. 

The term "Custodial Wallet" is independent of which LN Node is used. 
We should rethink this table and categorization.

**Which part needs editing?**
table Broad Categories of "LN Wallets"

**What editing does this need?**
Correction, clearer distinction, more precise distinction

**How does it currently read?**
see above

**How should it read?**
Here is a first draft, but this might not be perfect either, but at least this draft is consistent:

There are many ways wallets can be characterized or categorized. From all characteristics the two most important criteria are LN Node use and fund custody. Correspondingly the two most important questions to ask about a specific wallet are:
- Does this Lightning wallet have a full LN Node or does it use a 3rd-party LN Node?
- Does this Lightning wallet store its own keys or are the keys held be a 3rd-party custodian?

From these two questions we can derive 4 possible categories. However, the wallet category where a full LN node is used but the keys are held by a custodian is so rare that we ignore it from our discussion. That leaves us with three commonly found categories, but again remember that this is just one way of categorizing Lightning wallets.

[[lnwallet-categories]]
.Three commonly found categories of "LN Wallets"
|===
| Wallet Category    | LN Node        | Keystore/Custody | Technical Skills Required
|===
| Full-Node Non-Custodial Wallet | Full Node      | Non-Custodial | High
| 3rd-Party-Node Non-Custodial Wallet | 3rd-party node | Non-Custodial | Medium
| 3rd-Party-Node Custodial Wallet | 3rd-party node | Custodial     | Low
|===

[...]

PR #213, opened 2020-04-24
change wallet categories into wallet quadrant

8go:
The original categorization was not perfect. It was not ideal. It was sort of misleading. It sort of implies that a "Full Node & Wallet" is not a "Non-Custodial Wallet" because they are 2 different categories. It also states that a "Non-Custodial Wallet" uses "3rd-party node" for LN Node which is not the true definition.Things were getting mixed up here and we are not communicating a clear message.

The term "Custodial Wallet" is independent of which LN Node is used.
We should rethink this table and categorization.

I wrote an Issue about it #176 .

The Quadrant seems to be more useful and certainly more accurate.

Then I explained the terms "Lightweight" and "Node" because these terms were suddenly used, out of nowhere with no explanation.


02_getting_started.asciidoc
```diff
@@ -36,33 +36,48 @@ If users decide to outsource some functionality to a third party they usually gi
 
 Finally, those seeking simplicity and convenience, even at the expense of control and security, may choose a custodial Lightning wallet. This is the least technically challenging option, but it _undermines the trust model of cryptocurrency_ and should therefore be considered only as a stepping stone towards more control and self-reliance.
 
-In <<lnwallet-categories>> we see the three broad categories of lightning wallets and the relative degree of control they offer to the user.
+There are many ways wallets can be characterized or categorized. 
+From all features the two most important criteria are LN Node use and fund custody. Correspondingly the two most important questions to ask about a specific wallet are:
```
Roasbeef: I think a third question is something along the lines of where it gets its Bitcoin blockchain data from. Even if you have a "full" LN wallet, and control your own keys, if you blindly accept bogus data from a 3rd party, then you can actually be using "fake coins" without knowing it.

02_getting_started.asciidoc
```diff
@@ -36,33 +36,48 @@ If users decide to outsource some functionality to a third party they usually gi
...
+
+- Does this Lightning wallet have a full LN Node or does it use a third-party LN Node?
+- Does this Lightning wallet store its own keys under user control (self-custody) or are the keys held be a third-party custodian?
+
+From these two questions we can derive four possible categories. We can place these four categories into a quadrant as seen in Table <<lnwallet-categories>>. 
+But remember that this is just one way of categorizing Lightning wallets.
 
 [[lnwallet-categories]]
-.Broad Categories of "Lightning Wallets"
+.Lightning Wallets Quadrant
```
Roasbeef, line 52: I dig this matrix idea! I think we should add another row for the items I mentioned in #213.

02_getting_started.asciidoc
```diff
@@ -36,33 +36,48 @@ If users decide to outsource some functionality to a third party they usually gi
...
-| Wallet Type          | LN Node        | Keystore/Custody | Technical Skill
-| Full Node & Wallet   | Full Node      | Non-Custodial    | High
-| Non-Custodial Wallet | 3rd-party node | Non-Custodial    | Medium
-| Custodial Wallet     | 3rd-party node | Custodial        | Low
+|                        | *Full LN Node*      | *3rd-party LN Node*    
+| *Self-Custody*         | Q1: High technical skill | Q2: Medium technical skills 
+| *Custodial*            | Q3: not a common wallet type | Q4: Low technical skills 
 |===
 
+Q3, quadrant 3, where a full LN node is used but the keys are held by a custodian is not common. A user who goes through the effort of running a Ligtning Node will most likely also put in the additional efffort to store his own keys.
```
Roasbeef: True that it's not common, but it may become attractive in the future, as it would let a user worry about the operational aspects of their node, but then possibly delegate the access to the keys+signing to a 3rd party which may use primarily cold storage.

02_getting_started.asciidoc
```diff
@@ -36,33 +36,48 @@ If users decide to outsource some functionality to a third party they usually gi
...
+| *Custodial*            | Q3: not a common wallet type | Q4: Low technical skills 
 |===
 
+Q3, quadrant 3, where a full LN node is used but the keys are held by a custodian is not common. A user who goes through the effort of running a Ligtning Node will most likely also put in the additional efffort to store his own keys. 
+
 Lightning wallets can be installed on a variety of devices, including laptops, servers, and mobile devices. To run a full Lightning node (one that participates in "gossip" and creates its own map of the network for path finding and routing) you will need to use a server or desktop computer, as mobile devices and laptops are usually not powerful enough in terms of capacity, processing, battery life, and connectivity.
 
+The category "Third-party Ligntning Nodes" can again be subdivided into:
+
+- Lightweight: means that the Lightning Node is operated by a third party and that the wallet obtains the required information through an API that connects wallet and third-party LN node.
```
Roasbeef, line 67: Hmm, I think we're getting close here, but there's still more nuance. One example is the new Phoenix wallet. The user controls the keys, grabs blockchain data from Electrum servers (I believe, making them SPV in a sense), but delegates stuff like opening+managing channels and path finding to a 3rd party server. 

It sits somewhere in the middle, making certain tradeoffs to provide a better user experience in a mobile setting. 

Something on the far opposite end is something like Bluewallet, or other fully custodial wallets. In this case, the user has minimal control and trusts the operator of the hub node to not run away with funds. In the example of the Phoenix wallet, the user still maintains control of their keys.

02_getting_started.asciidoc
```diff
@@ -36,33 +36,48 @@ If users decide to outsource some functionality to a third party they usually gi
...
-| c-lightning    | Server  | Full Node   | Bitcoin Core          | User Control
-| Eclair Server  | Server  | Full Node   | Bitcoin Core/Electrum | User Control
-| Zap Desktop    | Desktop | Full Node   | Bitcoin Core/btcd     | User Control
-| Eclair Mobile  | Mobile  | Lightweight | Electrum              | User Control
-| Breez Wallet   | Mobile  | Full Node   | Bitcoin Core/btcd     | User Control
-| Phoenix Wallet | Mobile  | Lightweight | Electrum              | User Control
+| lnd            | Server  | Full Node   | Bitcoin Core/btcd     | Self-Custody
+| c-lightning    | Server  | Full Node   | Bitcoin Core          | Self-Custody
+| Eclair Server  | Server  | Full Node   | Bitcoin Core/Electrum | Self-Custody
+| Zap Desktop    | Desktop | Full Node   | Bitcoin Core/btcd     | Self-Custody
```
Roasbeef: For both Zap and Breez, we might want to add "neutrino" instead as the Bitcoin node category.

Roasbeef (changes requested):
Thanks for this PR! I agree we can improve the way we describe the various types of wallets, as there's a lot of nuance there, and as time goes on, developers are creating even newer models. I really dig the introduction of this matrix/quadrant system, as we can extend it in the future, and it gives readers a peek into all the combinations (and innovation!) that's possible.

02_getting_started.asciidoc
8go: Added the question. See Commit # 2  72314dab8d709783af287d97600e712b9dc91a06

02_getting_started.asciidoc
8go: I rephrased and worked your idea into the text. See Commit # 2.

See 72314dab8d709783af287d97600e712b9dc91a06

02_getting_started.asciidoc
8go: How is this not captured accurately in the table?
Phoenix Wallet | Mobile  | LN=Lightweight | BTC=Electrum | Self-Custody

Let's look at your sentence:
Phoenix wallet. The user controls the keys,  --> Self-custody

grabs blockchain data from Electrum servers (I believe, making them SPV in a sense),  --> Electrum

 but delegates stuff like opening+managing channels and path finding to a 3rd party server --> LN=3rd-party Lightweight

For Phoenix: As far as I can see your sentence matches exactly the table. What is it that I am not seeing?
Same for Blue Wallet. 
Unless I am missing something, it is OK as is. Please give further details to clarify. Thanks.

02_getting_started.asciidoc
8go: Currently:
| Zap Desktop    | Desktop | Full Node   | Bitcoin Core/btcd     | Self-Custody

I agree, this creates the wrong implication. It could lead to the incorrect conclusion that Zap Wallet runs a Bitcoin Core Full node. Not so!
This needs to be fixed. Thanks for catching that and pointing it out!

I fix Zap and Breez in Commit # 2.  72314dab8d709783af287d97600e712b9dc91a06

02_getting_started.asciidoc
8go: reworked Quadrant and text,
But I still kept it as a Quadrant. 
As high-level as possible-> 4 quadrants

See 72314dab8d709783af287d97600e712b9dc91a06

8go:
Look at 72314dab8d709783af287d97600e712b9dc91a06

I worked all comments in. I think the PR can be pulled.

The chapter has to be improved further over time. Many wallets will need to be added ....

This PR at least sets the structure and terms to use in table.

02_getting_started.asciidoc
```diff
@@ -36,33 +36,59 @@ If users decide to outsource some functionality to a third party they usually gi
...
+The category "Third-party Ligntning Nodes" can again be subdivided into:
+
+- Lightweight: means that the Lightning Node is operated by a third party and that the wallet obtains the required information through an API that connects wallet and third-party LN node.
+- None: means that not only is the Lightning Node operated by a third party but most of the wallet is operated by a third party in the cloud such that the wallet does not even need to make calls on a Lightning node API.
+
+These subcategories are used in Table <<lnwallet-examples>>.
+
+Other terms that need explanation in Table <<lnwallet-examples>> in column "Bitcoin Node" are:
+
+- Neutrino: Neutrino is a protocol to get data from Bitcoin. The Bitcoin Node is run by a third-party and accessed via "neutrino".
```
Roasbeef: Can point to BIP 157+158. Neutrino is the name of a _specific_ implementation of the client-side/wallet portion of these BIPs. However overtime, people have come to colloquially refer an implementation of the BIPs as "neutrino".

Roasbeef (changes requested):
One final correctness comment. Agree that this is just a start and will evolve a ton over time.

02_getting_started.asciidoc
8go: Good idea, that makes is better.

Done, see Commit # 3 in 6d3e4e37409994a2ebd2e1f35c53196a32d6ca4f

renepickhardt:
Thanks a lot!

PR #219, opened 2020-05-06
multi: add skeleton chapter outlines 

Roasbeef:
In this PR, we add outlines for prospective chapters. As the final chapter structure/ordering isn't finalized, they don't include an explicit chapter number. Each chapter outline includes a high level overview, as well as a series of questions that the reader should be able to answer after reading the chapter. These questions are also meant to server as a guide to prospective contributors as well. Some of the questions posted have already been answered in existing chapters checked into the repo.


Roasbeef:
Easier to read in raw text vs rendered ascii doc...

PR #224, opened 2020-05-12
Fix 'loose' vs 'lose' mistakes

oneeman:
Saw a place where it says "loose" when it should be "lose".  Searched through the project to find other places with the same mistake and fixed them.


oneeman:
Just noticed those chapters are marked as "EARLY DRAFT - In progress, changing often: Submit issues only, NO PRs or fixes".  Still, for such trivial fixes maybe a PR is appropriate, sorry for the trouble, and please do as you see fit.

bolatovumar (approved):

PR #253, opened 2020-06-14
maliciously

HugoDoyon:
maliciously instead of malicously


HugoDoyon:
This one is a simply a typo

03_how_ln_works.asciidoc
```diff
@@ -524,10 +524,11 @@ It is important to understand that HTLCs have a time measured in absolute blockh
 Once the sender of a payment sends away the onion it is completely out if their control what happens.
 Honest nodes SHOULD either forward the onion as quickly as possible or send an error back the original.
 While sender expects nodes along the path to be honest it has according to the protocol no power of making sure that nodes act quickly.
-Thus payment can currently get stuck if nodes want to behave malicously. 
+
+
+Thus payment can currently get stuck if nodes want to behave maliciously.
```
renepickhardt: shouldn't payment also be be `payments` in plural?

renepickhardt (commented):
just found another typo in that sentence would prefer to fix both so that we don't forget before merging

PR #255, opened 2020-06-14
initiate

HugoDoyon:
initate should be initiate


aantonop:
FYI, if you are making a lot of single-word and typo corrections in the same file, you can do those as multiple commits in one pull request, you do not need to do one pull request per commit. In fact, it would be better if you consolidated, since github's interface for merging PRs is clunky

Thanks

HugoDoyon:
Well noted will do, thank you Andreas!

PR #256, opened 2020-06-15
glossary: minor cleanup, consistent naming, etc

8go:
- typo
- consistent naming:
  - uppercased all "lightning" to "Lightning"
  - all Lightning networks, to "Lightning Network"
- remove "oldid" URL REST parameter from wikipedia links
- added links where appropriate
- etc


renepickhardt (commented):
the oldid REST links where there on purpose because they fix the version that I have read at the time of citing the article. We should have a general decision how we want to quote wikipedia links @Roasbeef @aantonop

8go:
OK, I didn't know. 

I put the oldid back in as they were before, this way you can close this PR. 
I can open up another PR, just with the oldid's and nothing else. 
This way we can move forward more quickly. 

IMHO: latest version is preferred as one has to assume that Wikipedia is improving over time (and not getting worse)

renepickhardt (approved):

PR #257, opened 2020-06-15
ch03: Scarcity, scarce resource differences between Bitcoin and Lightning

8go:
- added additional difference between BTC and LN

### DIFF
--- a/03_how_ln_works.asciidoc
+++ b/03_how_ln_works.asciidoc
@@ -721,9 +721,9 @@ When settled on-chain, it must be settled in "satoshis".
 Bitcoin and Lightning have different scarce resources. 
 The scarce resource for operating on the Bitcoin protocol is the space in the blockchain. 
 The scarce resource for operating on Lightning is the channel capacity, or in other words the liquidity.
-Hence, the two different pricing models, each one according to its scarcity. 
-Bitcoin transaction fees pay for block space. 
-Lightning transaction fees pay for channel capacity. 
+Hence, the two different pricing models, each one influenced by its scarcity. 
+Bitcoin transaction fees pay primarily for block space. 
+Lightning transaction fees pay by and large for channel capacity. 
 
 === Commonality of Bitcoin and Lightning
 
--- a/03_how_ln_works.asciidoc
+++ b/03_how_ln_works.asciidoc
@@ -721,9 +721,6 @@ When settled on-chain, it must be settled in "satoshis".
 Bitcoin and Lightning have different scarce resources. 
 The scarce resource for operating on the Bitcoin protocol is the space in the blockchain. 
 The scarce resource for operating on Lightning is the channel capacity, or in other words the liquidity.
-Hence, the two different pricing models, each one influenced by its scarcity. 
-Bitcoin transaction fees pay primarily for block space. 
-Lightning transaction fees pay by and large for channel capacity. 
 
 === Commonality of Bitcoin and Lightning
 
--- a/03_how_ln_works.asciidoc
+++ b/03_how_ln_works.asciidoc
@@ -717,6 +717,14 @@ Lightning nodes can optionally be configured to work with "milli-satoshis".
 So, within Lightning even smaller amounts can be exchanged. 
 When settled on-chain, it must be settled in "satoshis".
 
+==== Blockchain Space as Scarcity vs. Channel Capacity as Scarcity
+Bitcoin and Lightning have different scarce resources. 
+The scarce resource for operating on the Bitcoin protocol is the space in the blockchain. 
+The scarce resource for operating on Lightning is the channel capacity, or in other words the liquidity.
+Hence, the two different pricing models, each one according to its scarcity. 
+Bitcoin transaction fees pay for block space. 
+Lightning transaction fees pay for channel capacity. 
+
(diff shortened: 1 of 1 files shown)


03_how_ln_works.asciidoc
```diff
@@ -717,6 +717,14 @@ Lightning nodes can optionally be configured to work with "milli-satoshis".
 So, within Lightning even smaller amounts can be exchanged. 
 When settled on-chain, it must be settled in "satoshis".
 
+==== Blockchain Space as Scarcity vs. Channel Capacity as Scarcity
+Bitcoin and Lightning have different scarce resources. 
+The scarce resource for operating on the Bitcoin protocol is the space in the blockchain. 
+The scarce resource for operating on Lightning is the channel capacity, or in other words the liquidity.
+Hence, the two different pricing models, each one according to its scarcity.
```
renepickhardt: Not sure if I agree with the conclusion in respect to pricing of fees. In particular as the mempool in bitcoin is publically available and people can react the battle for space quickly. the liquidity in channels changes quickly and is not announced. 

Also if you'd have a high capacity channel following your argument the routing fees should be low because there is always liquidity. I believe others might argue that the price should be high because it is a reliable channel that offers premium routing services.

renepickhardt (commented):
While I like the scarcity comparision and also the fee comparision I am not sure if we should draw a connection between scarcity and fees. Seems more like an opinion to me. maybe a plausible one. but again I am not sure

03_how_ln_works.asciidoc
8go: Please re-read. There is no talk of "price" or "pricing" per se. It just states plainly what one pays for. It also does not imply when/if prices are low/high. It does not give an opinion of how prices will evolve or be computed. It just says: if you use X (Lightning) you pay for Y (e.g. capacity). 

Independent of this PR, just to chat or have a philosophical discussion, my opinion is: responding to your example: if there is only 1 provider of a high capacity channel this unique channel provider can ask for a fee premium. However, if the liquidity in general is high, i.e. if there are 1000 providers all providing that high capacity, then there will be competition and it will be difficult for 1 channel provider to ask for a fee premium. So, one has to look at *OVERALL* liquidity and how one capacity relates to others when one is interested in creating a pricing scheme. But anyway, the PR does not go into this detail or into pricing. It only tries to state what the user pays for under the hood.

8go: Please re-read. There is no talk of "price" or "pricing" per se. It just states plainly what one pays for. It also does not imply when/if prices are low/high. It does not give an opinion of how prices will evolve or be computed. It just says: if you use X (Lightning) you pay for Y (e.g. capacity). 

But I agree it is simplified. One pays for multiple things: electricity, compute power, but above all one pays for liquidity. I will add "primarily" or similar to PR to soften it a tiny bit.

Independent of this PR, just to chat or have a philosophical discussion, my opinion is: responding to your example: if there is only 1 provider of a high capacity channel this unique channel provider can ask for a fee premium. However, if the liquidity in general is high, i.e. if there are 1000 providers all providing that high capacity, then there will be competition and it will be difficult for 1 channel provider to ask for a fee premium. So, one has to look at *OVERALL* liquidity and how one capacity relates to others when one is interested in creating a pricing scheme. But anyway, the PR does not go into this detail or into pricing. It only tries to state what the user pays for under the hood.

8go: Softened in commit # 2 in d346aa8bf2d437e86c1a35b7bd46888c1e0eeb40
to weaken link between scarcity and fees

03_how_ln_works.asciidoc
```diff
@@ -721,9 +721,9 @@ When settled on-chain, it must be settled in "satoshis".
 Bitcoin and Lightning have different scarce resources. 
 The scarce resource for operating on the Bitcoin protocol is the space in the blockchain. 
 The scarce resource for operating on Lightning is the channel capacity, or in other words the liquidity.
-Hence, the two different pricing models, each one according to its scarcity. 
-Bitcoin transaction fees pay for block space. 
-Lightning transaction fees pay for channel capacity. 
+Hence, the two different pricing models, each one influenced by its scarcity. 
+Bitcoin transaction fees pay primarily for block space. 
+Lightning transaction fees pay by and large for channel capacity.
```
renepickhardt: I am not even sure if I understand the grammer of the addition `by and large`

renepickhardt (changes requested):
while it is not such strong language anymore I am still not sure in principle if the fees for lightning are there because of the scarcity of liquidity. With bitcoin and blockchain it seems obvious to me. In lightning I believe the situation is less clear. I would probably prefere avoiding the last three sentences. no need to do speculations in this book which is more about technology than economy anyway

03_how_ln_works.asciidoc
8go: This is just proper English. "by and large" is identical in meaning to "primarily". A synonym. See https://en.wiktionary.org/wiki/primarily.

8go: Took out the last 3 lines. As you said it is more about economy than about technology.

See commit # 3 in 74cdd1f0d47cef9b25ad884d851ecadd8ed1f00b

PR #260, opened 2020-06-15
ch03: Added monetary limit to commonality of BTC and Lightning

8go:
As an alternative a new paragraph could be added with title "Monetary Limit" but I thought that adding two sentences to the "Monetary Unit" section would be enough.

### DIFF
--- a/03_how_ln_works.asciidoc
+++ b/03_how_ln_works.asciidoc
@@ -724,7 +724,7 @@ When settled on-chain, it must be settled in "satoshis".
 Both the Bitcoin network and the Lightning network use the same monetary units: bitcoins. 
 Lightning payments use the very same bitcoins as Bitcoin transactions.
 As an implication, because the monetary unit is the same, also the economic monetary limit is the same: 21 million bitcoins.
-From these 21 million bitcoins, some are locked on the Bitcoin blockchain so that they can be used on Lightning.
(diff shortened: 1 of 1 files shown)


03_how_ln_works.asciidoc
```diff
@@ -723,6 +723,8 @@ When settled on-chain, it must be settled in "satoshis".
 
 Both the Bitcoin network and the Lightning network use the same monetary units: bitcoins. 
 Lightning payments use the very same bitcoins as Bitcoin transactions.
+As an implication, because the monetary unit is the same, also the economic monetary limit is the same: 21 million bitcoins.
+From these 21 million bitcoins, some are locked on the Bitcoin blockchain so that they can be used on Lightning.
```
renepickhardt: I would argue all bitcoins are always locked on the bitcoin blockchain. 

I think the more precise statement would be. "some are locked in 2 out of 2 multisig addresses, so that they can empower / enable a payment channel on the Lightning Network"

renepickhardt (changes requested):
please refine the sentence along the suggested idea

03_how_ln_works.asciidoc
8go: Adjusted to reflect your suggestions in commit #2 
a99bcf77bb088f045da575d5d96edd84b3d58997

8go:
change request completed.

renepickhardt (approved):

PR #261, opened 2020-06-15
ch01: Acronym LN (mostly) removed

8go:
- I removed acronym from first sentence and put it into the second. I did not feel that it is a good start when the first sentence of the book already has an acronym.
- Removed LN acronym 4-5 times further down. It was not consistently used anyway. Sometimes yes, sometimes no. I also think it reads better spelled out. So, for a book I feel it is better to spell it out 4 times, than to abbreviate it 4 times.

### DIFF
--- a/01_introduction.asciidoc
+++ b/01_introduction.asciidoc
@@ -2,13 +2,13 @@
 [[intro_what_is_the_lightning_network]]
 == Introduction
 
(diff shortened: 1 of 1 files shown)


renepickhardt:
I agree with that. but we had that discussion several times in the past 
@aantonop WDYT?

renepickhardt (approved):

PR #263, opened 2020-06-15
ch02: added website check to avoid going to fake website

8go:
- added 1 sentence: She verifies that the webpage is secure (https not http) which is shown as a green lock by some browsers.


renepickhardt (approved):
ACK

however I want to say that these kind of security checks are always limited anyways. At the time of reading the book someone could have hacked the ACINQ website or other circumstances might have emerged. 

Maybe we add another NOTE box about the limitations and general considerations about security?

8go:
I think that this location (ch02) is *not* the right place to put such a general note. She has been presented to the reader as a beginner, and the simple checks that a beginner can easily do are mentioned. 

Such a note should be put somewhere else in the book as a placeholder for the *WHOLE* book.

PR #264, opened 2020-06-18
glossary: "oldid" removed from Wikipedia URL links

8go:
This was already mentioned in PR #256 

@renepickhardt commented the following there: 
  > the oldid REST links where there on purpose because they fix the version 
  > that I have read at the time of citing the article. 
  > We should have a general decision how we want to quote wikipedia links
  @Roasbeef @aantonop

IMHO: I would remove oldid-s assuming that pages get better over time.

### DIFF
--- a/glossary.asciidoc
+++ b/glossary.asciidoc
@@ -37,7 +37,7 @@ Asymmetric Cryptographic System::
     Asymmetric cryptography, or public-key cryptography, is a cryptographic system that uses pairs of keys: public keys which may be disseminated widely, and private keys which are known only to the owner.
     The generation of such keys depends on cryptographic algorithms based on mathematical problems to produce one-way functions.
     Effective security only requires keeping the private key private; the public key can be openly distributed without compromising security.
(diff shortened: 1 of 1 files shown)


renepickhardt (commented):
While I agree that wikipedia articles tend to improve over time I believe in the context of writing an educational book it makes sense to reference to an article in a fixed form. If a user wants to know the current form they can check it out. but what could happen is that knowledge fundamentally changes due to new insights and then it would be strange if we quote something with wikipedia and in the article it is different. in that case it seems much more resonable to point to the version from which I got the information. 

Thus I would stick with oldids's even though they are ugly. I believe we will use an URL shortening sevice later anyway so for the user it is not so much about typing so much in

aantonop:
We will need to replace long links like this with short links, wherever possible. For readers of the print edition, it is not possible to transcribe links that are this long. In the past, we used a bit.ly shortener with an appendix showing the long and short versions in case bit-ly stopped working. We will probably do the same in post-production for this book. 

For now, the link issue is not important. If we use short links we will want to preserve the Wikipedia oldid version.

PR #269, opened 2020-06-18
ch02: global replacement of "LN Channel" with "Lightning Channel"

8go:
- see similar PR #268 
- 5 occurrences
- wallets already use the term "Lightning Channel" (instead of Lightning Network Channel)
- see 3 image of mobile wallet used in this book, all 3 images have term "Lightning Channel"


renepickhardt:
While I am in favor of naming things `Lightning Thing` instead of `LN thing` I believe in this particular case we should just say `payment channel `

8go:
> While I am in favor of naming things `Lightning Thing` instead of `LN thing` I believe in this particular case we should just say `payment channel `

- That is a fair and valid alternative.
- What do other contributors think/prefer?
- Some wallets already use term "Lightning Channel".

8go:
- I suggest committing/accepting this PR now because "Lightning Channel" is better than "LN Channel". So, we moved forward a step, we improved. 
- If later the general consensus is that "Payment Channel" is better than "Lightning Channel" then I would (later) adjust the whole book to use "Payment Channel" instead of "Lightning Channel".

PR #283, opened 2020-06-24
channel construction: pt4: English, etc

8go:
- To repeat what you should already know : ==> might offend reader who does not remember, etc. ==> eliminated
- established std writing is 2-out-of-2 
- I tried to highlight the first (and only the first) occurrence of the terms established in chapter 03 (e.g. "commitment transaction")
- secrete ==> secret
- revokation ==> revocation
- etc


aantonop:
@8go This chapter/section is still in "Drafting" status and is not ready for PRs or review. It is still very much in flux. PRs will create conflicts, since the underlying text will be changing a lot, requiring a lot of de-conflicting work. I will close these PRs for now. 

Please see the README for information about which chapters are ready for review.

PR #293, opened 2020-07-01
tiny English improvements

8go:
- Similarily ==> correspondingly   ... fits the meaning better
- show ==> present ... more elegant
- BOLTS ==> BOLT standards ... to make it more clear to the novice reader
- removed comma ... no comma when subject is same in both parts of sentence
- recommend ==> endorse ... better, recommend one over the other is at best used colloquially

### DIFF
--- a/node_client.asciidoc
+++ b/node_client.asciidoc
@@ -1,17 +1,17 @@
 [[set_up_a_lightning_node]]
 == Lightning Node Software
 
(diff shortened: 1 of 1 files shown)


aantonop:
I now see that I said I would present the clients in alphabetic order, but ended up writing LND before Eclair. I will need to reorganize a bit...

PR #308, opened 2020-07-06
Section: Why is it important to keep the packet fixed sized at all times?

ImranLorgat:
Wrote up this section, describing how the onion layers are peeled as it moves across the route, as well as giving examples of how malicious entities could potentially use packet size to spy on and de-anonymize users


ImranLorgat:
This is the diagram referred to in the following commit:

https://github.com/lnbook/lnbook/pull/308/commits/06bb131b1d05120e593b9c689070f1113e869759

malicious-routing-diagram

onion-routing-htlc-forwarding.asciidoc
```diff
@@ -19,3 +19,48 @@ Relevant questions to answer:
...
+
+We can extend this example.
+Imagine a malicious entity sets up multiple Lightning nodes connected to other well-connected nodes, and it also connects itself across popular payment routes.
+These malicious Lightning nodes would then be popular routing choices for those wanting to send payments, especially if they set their routing fees low.
+The malicious nodes can then capture the data of all packets that pass through their routing nodes.
+With additional information, such as the names of the other routing nodes, it could infer who is making these payments, who is receiving them, and for what amounts.
+footnote:[Note that not all Lightning nodes are anonymous.
+It is known, for example, that the nodes "aantonop" and "1.ln.aantonop.com" are owned by the author of this book Andreas Antonopolous.
+Furthermore, many companies and businesses in this space use a publicly known name for their routing nodes.
+If we see, for example, a payment with destination "Bitrefill", we could infer that someone is making a purchase from Bitrefill.
```
renepickhardt: actually we can't as names are just set by the node operator and not assigned by the domain name system. thus anyone can name their node as they wish and we can't really deduce anything from the  `alias`

onion-routing-htlc-forwarding.asciidoc
```diff
@@ -19,3 +19,48 @@ Relevant questions to answer:
...
+If we know the prices of their services, we could even infer what they purchased. ]
+If it has multiple routing nodes connected to each other, it might even be responsible for transmitting several of the hops in a single payment and could form a more complete picture of the route.
+We see this example as spiritually similar to the chain analysis already performed on the Bitcoin network; even an incomplete picture of payments can be used to infer things about the parties involved and potentially de-anonymize them.
+
+Fixing the packet size solves the problem of routing nodes knowing how far a packet is along the route.
+Whenever a routing node peels off a layer, it adds another layer of junk data to the end, restoring the packet to its "full" size.
+In this way, even if a packet has already been transmitted 18 hops, the 19th routing node would still see that the packet contains enough data for up to 20 hops.
+It would have no way of determining if it was the first routing node or the second last routing node.
```
renepickhardt: no way is quite a strong statement... 

there are other methods that a routing node can use to infer this information and at least drastically reduce the candidate set

renepickhardt (changes requested):
I am in favor of this PR. though I would resolve the two comments. 

I also suggest to change the names in the diagrams to mallory and another of our names for the second last node and potential destination nodes. 

Do you have the source file of your diagram somewhere?

onion-routing-htlc-forwarding.asciidoc
ImranLorgat: Thanks for the review Rene, I've worded it better in the commit below to include node pubkeys. You're right, I could spoof you myself by naming my node "renepickhardt". But you could claim ownership of your own node by posting on Twitter: "I own a node with X alias and Y pubkey", in which case your node now becomes "known"

https://github.com/lnbook/lnbook/pull/308/commits/ce45e40d2439898067d819b07a39d6f71ce81375

onion-routing-htlc-forwarding.asciidoc
ImranLorgat: Got you. Softened the language here:

https://github.com/lnbook/lnbook/pull/308/commits/a316125e50536194ccfee934af6167c0a08ca551

ImranLorgat:
> Do you have the source file of your diagram somewhere?

I am a pleb so I made it in Microsoft Word. Not sure who should be Mallory or the other names, so happy to upload and for you to change it as you wish. I assume closer to the final stages you'll have someone propose a prettier looking diagram than this.

[Onion Routing Diagram.docx](https://github.com/lnbook/lnbook/files/4996514/Onion.Routing.Diagram.docx)

onion-routing-htlc-forwarding.asciidoc
renepickhardt: agree and I think ok for merging. In general we try to describe the protocol. so what people do outside of it is kind of their problem. of course in general people act outside of the protocol. --> ACK

onion-routing-htlc-forwarding.asciidoc
renepickhardt: much better (:

renepickhardt:
thanks for the file (: yes diagrams will be remade by the publisher anyway. but still we need to provide them input. I agree this one we could have easily made ourselves but in general it is good to have the source files.

PR #309, opened 2020-07-06
Section: Source-Based Routing

ImranLorgat:
Wrote an explanation for Source-Based Routing, why it is used and other methods are not, and some words on the drawbacks of using it

### DIFF
--- a/path-finding.asciidoc
+++ b/path-finding.asciidoc
@@ -25,7 +25,7 @@ As discussed in the chapter on Onion Routing, the intermediary nodes transmittin
 As such, intermediary nodes cannot assist with path-finding because this requires them to know the destination of the payment.
 Without knowing the destination, they cannot calculate whether the path selected is efficient or not.
 
-We also cannot expect the destination to find a path.
+We also cannot expect the destination to find a path, although it can provide routing hints that can assist the source in finding one.
 Like Bitcoin, the Lightning Network uses a "push" system to make payments.
 In Bitcoin, payments are pushed by the sender to the receiver's address, and in Lightning, payments are pushed from the sender's channel to fulfill the receiver's invoice.
 The recipient is unable to "pull" a payment from anyone.
--- a/path-finding.asciidoc
+++ b/path-finding.asciidoc
@@ -22,7 +22,7 @@ While a payment is in-flight along a path, the path typically does not get chang
 
 
 The Lightning Network uses source-based routing at the time of writing in order to protect user privacy.
-As discussed in the chapter on Onion Routing, the intermediary nodes transmitting the payment are not aware of the full route of the payment; they only know the node they received it from and the node they are sending it to.
+As discussed in the chapter on Onion Routing, the intermediary nodes transmitting the payment are not aware of the full path of the payment; they only know the node they received it from and the node they are sending it to.
 As such, intermediary nodes cannot assist with path-finding because this requires them to know the destination of the payment.
 Without knowing the destination, they cannot calculate whether the path selected is efficient or not.
 
--- a/path-finding.asciidoc
+++ b/path-finding.asciidoc
@@ -17,7 +17,7 @@ Relevant questions to answer:
 === What is "Source-Based" routing and why does the Lightning Network use it?
 
 Source-based routing is a method of path-finding where the sender (i.e. the source) plans the route from itself, through the intermediary nodes, and finally to the destination.
-Once a route has been selected, the sender sends the payment to the first intermediary node, who sends it to the second intermediary node and so on, until it reaches the destination.
+Once a path has been selected, the sender sends the payment to the first intermediary node, who sends it to the second intermediary node and so on, until it reaches the destination.
 While a payment is in-flight along a route, the route cannot be changed by any of the intermediary nodes, even if a shorter path or a cheaper path (in terms of routing fees) exists.
 
 The Lightning Network uses source-based routing at the time of writing in order to protect user privacy.
--- a/path-finding.asciidoc
+++ b/path-finding.asciidoc
@@ -31,7 +31,7 @@ Even if it specifies a path in the invoice, that path may no longer be viable by
 The recipient can, however, specify "routing hints" in the invoice that may assist the sender in finding a possible path.
 
 Furthermore, source-based routing comes with some inherent drawbacks.
-The sender chooses the route based on their current understanding of the topological map of the Lightning network.
+The sender chooses the path based on their current understanding of the topological map of the Lightning network.
 As discussed in previous chapters, this map is necessarily incomplete; the sender may not be aware of all the channels, and even if they are they will almost certainly not know the latest balances in each of the channels.
 And even if the sender did have a complete topological map at one point in time, the balances of channels change with every payment, and so in a short space of time the map would become obsolete.
 It is for this reason that the sender probes various paths until it finds a satisfactory path, and then sends the payment.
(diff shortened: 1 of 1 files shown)


path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
   * What information can be sent to intermediate and the final node aside from the critical routing data?
   * What are multi-hop locks? What addition privacy and security guarantees to they offer?
   * How can the flexible onion space be used to enabled packet switching in the network?
+
+
+=== What is "Source-Based" routing and why does the Lightning Network use it?
+
+Source-based routing is a method of path-finding where the sender (i.e. the source) plans the route from itself, through the intermediary nodes, and finally to the destination.
+Once a route has been selected, the sender sends the payment to the first intermediary node, who sends it to the second intermediary node and so on, until it reaches the destination.
```
renepickhardt: nit: once a PATH has been selected...

path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
...
+While a payment is in-flight along a route, the route cannot be changed by any of the intermediary nodes, even if a shorter path or a cheaper path (in terms of routing fees) exists.
```
renepickhardt: ...along a path, the path cannot... 

though not sure if that is true. the current system technically allows for JIT routing which technically changes the path to some degree. at least it seems equivalent to link level AMP

path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
...
+
+The Lightning Network uses source-based routing at the time of writing in order to protect user privacy.
+As discussed in the chapter on Onion Routing, the intermediary nodes transmitting the payment are not aware of the full route of the payment; they only know the node they received it from and the node they are sending it to.
```
renepickhardt: ...full path ...

path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
...
+As such, intermediary nodes cannot assist with path-finding because this requires them to know the destination of the payment.
```
renepickhardt, line 26: again not sure if this is true because the current system allows techniques like JIT routing

path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
...
+Without knowing the destination, they cannot calculate whether the path selected is efficient or not.
+
+We also cannot expect the destination to find a path.
```
renepickhardt, line 29: Though it can provide routing hints

path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
...
+Like Bitcoin, the Lightning Network uses a "push" system to make payments.
+In Bitcoin, payments are pushed by the sender to the receiver's address, and in Lightning, payments are pushed from the sender's channel to fulfill the receiver's invoice.
```
renepickhardt: to simplistic and semantically maybe not fitting in the source based routing section? 

In lightning HTLCs are set up from sender to receiver along a path of channels but once the path of htlcs is set up the recipient actually pulls the payment by setteling the htlcs which propagates back by a chain of reimbursments all the way to the sender

path-finding.asciidoc
```diff
@@ -12,3 +12,31 @@ Relevant questions to answer:
...
+The recipient is unable to "pull" a payment from anyone.
+Even if it specifies a route in the invoice, that route may no longer be viable by the time the invoice is paid, which could be several minutes or several days later.
+The recipient can, however, specify "routing hints" in the invoice that may assist the sender in finding a possible path.
+
+Furthermore, source-based routing comes with some inherent drawbacks.
+The sender chooses the route based on their current understanding of the topological map of the Lightning network.
+As discussed in previous chapters, this map is necessarily incomplete; the sender may not be aware of all the channels, and even if they are they will almost certainly not know the latest balances in each of the channels.
+And even if the sender did have a complete topological map at one point in time, the balances of channels change with every payment, and so in a short space of time the map would become obsolete.
+It is for this reason that the sender probes various paths until it finds a satisfactory path, and then sends the payment.
```
renepickhardt, line 37: probing and sending is kind of equivelent. the user could concurrently probe various paths with various fake payment hashes and then send along a path that seemed to have enough liquidity

renepickhardt (changes requested):
I would merge this to get started on this chapter but later in editorial change quite a bit. in particular stick to our definition / differences between path and route. actually we do not use the term route but we use path finding and routing.

path-finding.asciidoc
ImranLorgat: The nit has been picked. See below. Where can I find a reference for the difference between a path and a route? 

https://github.com/lnbook/lnbook/pull/309/commits/33a821d5477dc9303dbd2e60a94016959087b267

path-finding.asciidoc
ImranLorgat: Added

https://github.com/lnbook/lnbook/pull/309/commits/008c1f0f341ec2bed03c4435c794778f152da7e5

path-finding.asciidoc
ImranLorgat: Amended. Hmm, in my understanding JIT routing is a different method to source-based routing. However, I'm certainly not well-versed in JIT routing so happy for you to suggest an alternate explanation here.

https://github.com/lnbook/lnbook/pull/309/commits/47491e4c166c9f9169a0d90ef3437c0857415c85

path-finding.asciidoc
ImranLorgat: Oooook

https://github.com/lnbook/lnbook/pull/309/commits/6b053026fe2ed82123c77028865a8bde05dea2d4

path-finding.asciidoc
ImranLorgat: Hmmm.... not sure how to word it in this case. Happy for you to suggest an alternative or should I just kill these few lines?

path-finding.asciidoc
ImranLorgat: Yeah you're right, I oversimplified it a little too drastically. Decided to just kill those two lines. BTW looks like I did specify that the destination can provide routing hints below that line so removed the first mention of it.

https://github.com/lnbook/lnbook/pull/309/commits/56a7b75c0d6627f52f4f1bf49329b7c9ab21d5ff

path-finding.asciidoc
ImranLorgat: Hmm explain? I'm not too familiar with the mechanics of probing. I'm currently under the impression that you basically ask a bunch of nodes about their capacity and then try to figure out if the payment will work. Not sure how the fake payment hashes work. Could you suggest a more precise wording?

https://github.com/lnbook/lnbook/pull/309/commits/eed081a19fc3746ec0168bb207aea7e591869ebe

ImranLorgat:
Wait, did you _invent_ JIT routing on the Lightning Network? See link below. If so... badass...

https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-March/001891.html

path-finding.asciidoc
renepickhardt: I will merge and afterwards just delete those lines here. I think for now this is the more elegant way

path-finding.asciidoc
renepickhardt: in mastering the lightning network (: it is still an ongoing battle between terminology but at least we have agreed to make this distinction. and books in general are a good place to set terminology

path-finding.asciidoc
renepickhardt: I think for here this is much better. the telescoping will be explained elsewhere (hopefully / probably)

renepickhardt:
yeah I did. everyone nagged me about rebalancing and when I decided to create a rebalancing plugin for c-lightning I realized that JIT routing is a thing. it is actually funny how it turns out that it has many very good properties with respect to improve privacy of participants

PR #315, opened 2020-07-10
Update ch1 - minor phrasing

keblek:
This is a minor edit that to my ear did not sound correct


keblek:
I edited this by pressing the pencil in the top right corner of the document when viewing it in github. 

For quick and minor edits this workflow should be encouraged in the contributing guidelines, thoughts?

PR #319, opened 2020-07-10
code, spelling and grammar fixes for node_client chapter

vnprc:
- Existing code assumes jq is installed on the local machine. To run jq inside docker use the 'bash -c' flag with double quotes.
- missing newline char causes apt-get install command to fail when installing c-lightning pre-req libraries
- github does not display markdown lists correctly unless list items are preceded by an empty line
- 'build' -> 'built' in several places in the text
- when building eclair wallet, 'mvn -V' is an invalid argument, leading to scary build failure messages. Instead use 'mvn -v'
- 'If the network is running correct' -> 'If the network is running correctly'

### DIFF
--- a/code/docker/setup-channels.sh
+++ b/code/docker/setup-channels.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 echo Getting node IDs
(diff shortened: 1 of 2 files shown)


aantonop:
Excellent work!

PR #290, opened 2020-06-29
WIP: chapter 03 AMA edits

aantonop:
Edits on chapter 3, for language and clarity. 

This is Work-In-Progress, will remain open until all of chapter 3 is done.

PR #323, opened 2020-07-20
Enumerate using asciidoc for better readability

grunch:


aantonop:
We're doing a lot of edits to chapter 3 in a WIP PR (#290). This will be lost in the conflicts. 

But, I think it is a good idea to enumerate using asciidoc, instead of static numbering. I would like to see this fix on all the chapters, as a best practice. 

Suggestion: Do this or similar PR on every finished chapter, once a chapter has completed the first round of edits and there are no open PRs like #290

grunch:
Great, I'll do that :+1:

PR #328, opened 2020-07-24
Fill storage section in node operations chapter

trigger-xyz:
I am starting with this small contribution :)

I am wondering if I should go into this level of detail or if it would be better to just have 2 simple sections for all hardware like : 
 - Minimum configuration
    * Storage : 1TB HDD
    * CPU : ...
 - Recommended configuration
    * Storage : 1TB SATA SSD
    * CPU : ...

### DIFF
--- a/node_operations.asciidoc
+++ b/node_operations.asciidoc
@@ -56,7 +56,8 @@ image::[images/raspiblitz.jpg]
   * Minimum Hardware requirements
   * CPU
   * RAM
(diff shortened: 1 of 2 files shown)


aantonop:
Thanks!

This chapter will change frequently, so if you do contribute more try to do it in larger continuous chunks (exactly like you did here) that I can fit into the bigger outline --  rather than in small changes in several places, that might cause merge conflicts. 

Nice!

PR #329, opened 2020-07-24
Update ubuntu version to 20.04

trigger-xyz:
The new Ubuntu LTS is now 20.04. I updated it in the "node client" chapter and in the docker files.
I also tested the whole chapter from scratch after doing the change, it is working fine !


aantonop:
Excellent work. Thank you @trigger67

PR #350, opened 2020-08-10
 routing chapter edits 

Roasbeef:


ImranLorgat:
Just realized, some overlap between this pull request and two others I submitted here:

1. Language edits on the introduction, and created a diagram to illustrate
https://github.com/lnbook/lnbook/pull/369

2. Language edits on the gold coin to make it clearer
https://github.com/lnbook/lnbook/pull/370

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+However, as discussed earlier, they're compensated for this opportunity cost in the forms of earned routing fees when they successfully forward the payment.
 
-[NOTE]
-====
-Isn't it remarkable that even though the process of exchanging funds for an preimage seems to be happening concurrently at the same moment in time in reality it is actually happening one step after another but in the right order.
-====
+In the next section, we'll dive into the details of the "HTLC", exploring the properties that allow it to implement the concept of a conditional chained end to end secure payment _without_ involving the 3rd party escrow.
+You'll see how the Bitcoin network can be seen as a stand-in for a trusted third party to ensure proper HTLC contract execution. 
+After that, we'll explore how a user is able to use an HTLC to "route" a payment through the network securely.
+In the Lightning Network today, we use a technique called source-based onion routing, though it's also possible to route payment with alternative techniques.
```
renepickhardt, line 144: move a little bit up to the point where we talk about path finding

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
-In that case Bob could just force close the channel and publish his latest version of the commitment transaction.
-As the time lock of the HTLC is not over yet with an onchain success transaction Bob would be able to claim and settle his 15 mBTC as he is the only person who is able to spend the HTLC output in the commitment transaction.
-The other way around meaning Bob and Alice would negotiate a new commitment transaction with the removed HTLC would never be save for Alice.
-If the signatures for the new commitment transaction are exchanged Bob has received the money and could decide not to release the preimage.
+As long as the escrow is trustworthy and faithfully performs its duty, then no party will have their coins stolen in the process.
+The pre-condition to this _route_ working at all, is that all parties in the path already needed to have enough money to satisfy the required series of deposits.
+While this seems like a minor detail we will see in this chapter, this requirement is actually one of the harder issues for Lightning Network nodes.
```
renepickhardt, line 137: my feeling is that we should write "pathfinding" chapter as routing will be quite long and it makes sense to split those chapters

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+After that, we'll explore how a user is able to use an HTLC to "route" a payment through the network securely.
+In the Lightning Network today, we use a technique called source-based onion routing, though it's also possible to route payment with alternative techniques.
+Finally, we'll dive into the precise details concerning the exact mechanics of forward, settle ling, and cancelling HTLCs in the network.
+
+= Hash Time Locked Contracts as a Conditional Chained End to End Secure Payment =
+
+Our example in the prior section using "golden coins", was intended to lay same base intuition which we'll leverage in this section to explain how HTLCs work in practice.
+HTLC is actually an acronym that stands for "Hash Time-Locked Contracts". 
+A HTLC is a _specific_ instantiation of a Conditional Chained End to End Secure Payment (CCESP, don't use this acronym?).
+As we'll see in the later chapters, given a set of adequate cryptographic constructs, many other instantiations are possible as well.
```
renepickhardt, line 152: Would it make sense to give the coinbase transaction as an example as it depends on the condition that it solves the hash puzzle given the current difficulty. where should I put comments like this?

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+
+The payment should _only_ be able to be redeemed if a valid solution is revealed. Critical, all conditions need to be timed in order to allow the construct to return the funds back to the sender if a solution to this condition isn't revealed.
+The combination of the condition, and a timeout on the condition gives the payment a trait we commonly refer to as atomicity: either the payment happens, or the receiver if refunded the funds.
+
+=== Conditional Chained Payment ===
+
+Building upon our conditional payment, it may be possible to *chain* this payment, allowing it to involve the payer, the payee, and possibly several intermediaries. 
+Each intermediary, is able to present a _slightly_ modified version of the condition (without invalidating it all together), and so so in an iterated manner until the conditional payment reaches the payee.
```
renepickhardt, line 171: remove one `so` in `so so`

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+Once it reaches the payee, then the payment should be able to be _iteratively_ resolved, starting at the payee all the way back to the payer. 
+
+Each chaining creates an "incoming" and "outgoing" conditional payment.
+A node receives a conditional payment from a party (incoming condition), and then extends the conditional payment to the next party in the chain (outgoing condition).
+The payment is extended in from payer to payee, but settled from payee to payer, as each of the intermediaries gain the solution to the outgoing condition, and use that (possibly augmenting it) to satisfy the incoming solution.
+
+Typically the payer rewards the intermediaries by sending slightly more than the payment amount, in order to allow the intermediaries to send out less with their outgoing payment than what they received from the incoming payment.
```
renepickhardt, line 178: maybe this and the next sentence as an info box as it is a relevant idea but not necessary to understand conditional chained payments

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+Additionally, we also require that the amount the payer intended to send is fully received by the payee.
+Finally, we require that non of the intermediaries are able to "contaminate" the payment beyond giving incorrect directions to the party that directly follows them.
+In other words, the intermediary shouldn't be able to materially affect the propagation of the payment several hops away from it.
+
+== Hash Time Locked Contracts ==
+
+In this section, we'll construct a conditional chained end to end payment known as the HTLC. 
+At each step we'll add a new component, then examine it in light of our original definition to ensure all requirements and security properties are reached.
+
+First, the "condition". For an HTLC, the condition is typically the reveal of a hash pre-image that matches a particular hash.
```
renepickhardt, line 194: the following could be an ordered list. they are usually good for learners as it is like "5 steps to xyz"

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+   OP_CHECKLOCKTIMEVERIFY <timeout>
+   OP_DROP
+   <sender key>
+
+OP_CHECKSIGVERIFY
+```
+
+Alice can present this script to Bob in order to kick off the conditional payment.
+For the chained aspect, Alice needs to be able to communicate the proper payment details to each hop in the route.
+Recall that each hop will specify a forwarding fee rate, as well as other parameters that express their forwarding policy.
```
renepickhardt, line 218: lets split away the fee. I think I explain the fee in detail when I talk about the onion. Here (as above) it is kind of a distraction. It is correct but I think we don't loose anything if we leave that detail out

routing.asciidoc
```diff
@@ -82,304 +99,201 @@ As Wei gets message from the escrow that Bob has deposited the 10 golden coins W
...
+In addition to this forwarding rate, Alice also needs to be conceded about what time locks to use.
+Each node in the hop needs some time to be able to settle the outgoing, then incoming payment on-chain in the worst case.
```
renepickhardt, line 220: yeah time lock deltas are much more important AND we should mention in the script that we talk about an absolut timelock in blockheight (not everyone knows this when they read checlocktimeverify)

renepickhardt (changes requested):
Incomplete review about the abstract way of introducing HTLCs. 

Overall I like this very much. I have a few suggestions / wishes but I really like the level of abstraction.

routing.asciidoc
Roasbeef: Can you re-word this comment, I don't follow.

routing.asciidoc
Roasbeef: Hmm, not that's not the type of cryptographic construction I'm referring to here.

routing.asciidoc
Roasbeef: IMO it's a necessary piece of information for the "chained" aspects, as it's important that the incentives are built in and outlined.

routing.asciidoc
Roasbeef: The rest of the paragraph isn't a series of items though, unless you're referring to this section as a whole?

routing.asciidoc
Roasbeef: I agree we dive into it later in the chapter, but IMO it's important that the user is _introduced_ to the concepts earlier to serve as a base which we'l use to build on later in the chapter.

Roasbeef:
Merging this in preemptively as it's blocking some of the dependent changes, and the latter part of the chapter (including the diagrams) needs to be re-worked as it contains dated information.

PR #352, opened 2020-08-12
Add exchanges in chapter 2

ghost:
Lot of newbies find it difficult initially to swap BTC on chain to BTC on LN and vice versa. I have used some of the services/exchanges mentioned, they are helpful and it makes sense to include this list in chapter "getting started".

### DIFF
--- a/02_getting_started.asciidoc
+++ b/02_getting_started.asciidoc
@@ -113,6 +113,19 @@ In <<lnwallet-examples>> we see some examples of currently popular Lightning nod
 | Blue Wallet    | Mobile  | None        | None                  | Custodial
 |===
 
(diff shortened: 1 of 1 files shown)


renepickhardt (commented):
not sure. As the time of writing most of these services seem to be in a rather experimental stage

aantonop:
I think these might fit better in the node_operations chapter instead. What do you think @prayank23 ?

aantonop:
We added a whole section about adding liquidity and submarine swaps in the node_operations chapter. Closing this, but if you want to make changes to the node_operations sections to add a provider that you think is important, please do a new PR!

Thank you @prayank23

PR #367, opened 2020-08-25
Unknown participant in example

enorrmann:
The whole example talks about Alice and Bob... and suddenly it changes from Bob to Mallory ... copypaste error ?


renepickhardt (approved):
good catch

PR #371, opened 2020-08-27
reformatted + 1 bullet point

8go:
- reformatted for correct visualization
- added bullet point regarding chat apps using TLV

### DIFF
--- a/p2p.asciidoc
+++ b/p2p.asciidoc
@@ -1,44 +1,48 @@
 Chapter overview:
+
   * high level description of p2p interaction for the LN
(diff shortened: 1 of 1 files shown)


renepickhardt:
not sure if we really gonna talk about that point but the the bullet points are only suggestions for the moment we start writing anyway. so I will merge

8go:
@renepickhardt : Agreed. I am also unsure if this is the right place to mention chat apps. Maybe it will fit in better somewhere else. We will see.

PR #379, opened 2020-08-28
node_operations: English, minor adjustments/additions

8go:
- commas
- simplifications
- added downsides to running on laptop
- Backup is so crucial, I elevated it to full **...** formatting
- brought the figure of 300GB in June forward, today Aug 28 we are at 336GB, in Sept we will reach and exceed 340GB


aantonop:
Great work @8go thanks!

PR #389, opened 2020-08-29
Glossary - Private Channel

ImranLorgat:


aantonop:
WE might want to clarify that the term "private channel" is a misnomer. Unannounced channel is better since these channels are not private. They are included in routing hints and their existence and some information will leak when commitment transactions are recorded.

ImranLorgat:
@aantonop Noted. I was under the impression that they were known only to the channel partners themselves. With commitment transactions (and even funding transactions), this is expected since they're public.

Are "private"/"unannounced" channels discussed elsewhere in the book? I'll compare and update accordingly. Can't remember if they show up or not. Do you want to continue referring to them as "private", with a disclaimer, or refer to them as "unannounced"?

aantonop:
We need to clarify that these channels may be leaked by routing hints and commitment transactions. We can refer to them as private channels in the glossary, as that is the term that they are known by, but we should add as the first sentence that this is a misnomer and that "unannounced" is a better term for this concept.

Can you please edit accordingly? @ImranLorgat

PR #390, opened 2020-08-30
Glossary - Network Capacity

ImranLorgat:


aantonop:
Some typing errors "it the", please fix

ImranLorgat:
Fixed and reworded slightly for more clarity. Note, the definition mentions "private" channels, so this definition will need to be updated based on how we resolve the PR below:

https://github.com/lnbook/lnbook/pull/389

PR #392, opened 2020-08-30
How LN Works - Closing a channel (Language edits)

ImranLorgat:
Incorporated some edits from the defunct pull request below to make this section read a little more easily. Some particulars

* Instead of stating as good way, bad way, ugly way -> state as their technical name and put (the good way) bracketed after
* The first and third examples referred to "you" and the second referred to "Alice and Bob". All three now refer to "you"
* Each paragraph now describes the method with a one liner
* Some miscellaneous detail

https://github.com/lnbook/lnbook/pull/307

### DIFF
--- a/03_how_ln_works.asciidoc
+++ b/03_how_ln_works.asciidoc
@@ -263,7 +263,12 @@ When new nodes join the Lightning Network they collect the channel announcements
 
 ==== Closing the channel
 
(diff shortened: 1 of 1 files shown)


aantonop:
For consistency, should the heading for "Protocol breach" also include "(the ugly way)"?

ImranLorgat:
Oops. Slipped up. Fixed it.

PR #468, opened 2020-09-07
node_operations: backup, commas, rephrasing

8go:
- comma (no comma when or is used between 2 items)
- comma (use comma after "by default, ")
- comma (use Oxford comma for consistency when there are 3+ items in "and" or "or)
- an SCB (with "n")
- rephrasing the argument for SCBs


aantonop:
I really have to learn how to use commas correctly! thanks @8go

8go:
Unfortunately the comma rules are often very obfuscated. If it were easy, everyone would place them correctly. On top of that, there are even areas where the scholarly "experts" disagree. Don't worry. The content is *much* more important than the commas.

8go:
As I said I really don't see it as important, but if you are curious one day, here is a simple 1-pager:
https://www.grammarly.com/blog/comma/

PR #473, opened 2020-09-07
chapter_01: update proposal for LN definition and motivation

dcarpintero:
- updated LN definition and motivation.
- added footnote to Mastering Bitcoin reference.
- minor editorial changes.


aantonop:
This PR includes a lot of changes to sections that are both controversial and subjective. We are very carefully calibrating to avoid being too opinionated and also choosing our words carefully. While some of these changes are great, others would need a bit more discussion. Rather than trying to handle them all in one PR, I would suggest breaking this into separate PRs for each section. That way we can accept the easy ones quickly and discuss the rest. 

Please break into smaller PRs @dcarpintero

PR #484, opened 2020-09-16
Glossary - Payment Channels

### LINKED ISSUE #463: Fix needed: glossary: payment channels
**Describe the error**
This glossary entry refers to previous work on payment channels:

```
There are currently three methods known to construct a fully duplex bidirectional payment channel.
    Christian Decker proposed a method in his PhD thesis based on invalidation trees.
    In the Lightning Network whitepaper, Joseph Poon and Tadge Dryja describe the Revocable Sequence Maturity Contract based method that is currently being implemented on the Lightning Network.
    Recently Christian Decker et al. came up with the Eltoo mechanism which would require a Bitcoin softfork.
```

**To Fix**
A few suggestions:

* clarify that "a method based on invalidation trees" is the same as "Duplex channels", mentioned in the book previously.
* does a glossary entry even need such a detailed account of all existing constructions?
* if it does, Christian Decker's co-authors should probably also be mentioned (both Duplex and Eltoo papers have multiple authors)
* I'd rather suggest that the author names and details be omitted for brevity, and provided as footnotes or references: "there are three constructions: Duplex [1], Lightning [2], and eltoo [3]".
* aren't there more than three payment channel constructions? One could argue that the fourth one is [Spillman channels](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2013-April/002433.html), [implemented in BitcoinJ](https://bitcoinj.github.io/working-with-micropayments)). And the initial Satoshi's proposal (based on `nLockTime`, not enforceable) count?

ImranLorgat:
Fixes #463 

The glossary contains two separate definitions for payment channels, both of which are probably too complex for a glossary

### DIFF
--- a/glossary.asciidoc
+++ b/glossary.asciidoc
@@ -341,23 +341,18 @@ payment::
     The invoice includes a payment hash which is the hash of a secret preimage.
     This payment hash is used by the Hashed Time Lock Contracts during the routing process.
 
-payment channels::
-    A micropayment channel or payment channel is a class of techniques designed to allow users to make multiple Bitcoin transactions without committing all of the transactions to the Bitcoin blockchain. In a typical payment channel, only two transactions are added to the blockchain, but an unlimited or nearly unlimited number of payments can be made between the participants.
-
 payment channel::
-    Payment channels are the core building blocks of the Lightning Network.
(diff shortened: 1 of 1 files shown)


ImranLorgat:
Duplicate of #479. For some reason got opened twice.

PR #485, opened 2020-09-16
Glossary - Bitcoin Mining

### LINKED ISSUE #462: Editorial: style of glossary
**Which part needs editing?**
IMO, many entries of the glossary have the following shortcoming: they don't define the term directly, but rather compare it to some other terms, which the reader may not be familiar with. I suspect it would be easier for the reader if every entry has the form "X is Y", and then expands upon that if necessary. Some of the details can be omitted, or provided as references to other glossary entries.

Examples:

> **Balance**. The balance of a payment channel is encoded...

I'm expecting to first learn what the balance _is_, and only then how it is encoded. Suggestion: "The balance is the amount of coins a channel party currently controls. Balances are encoded... Balances change when... The sum of the two balances equals capacity (see: capacity)."

> **Bitcoin mining**. Bitcoin mining serves two purposes...

Again, what _is_ Bitcoin mining? Suggestion: "Bitcoin mining is a process of constructing a block from recent Bitcoin transactions and performing the required proof-of-work".

> **Neutrino**. Neutrino is a later alternative to SPV...

What's SPV then? An unfamiliar reader would have to go to another definition to understand this one. Suggestion: "Neutrino is a protocol that allows Bitcoin users verify that a transaction is included in a block without fully downloading this block. An earlier protocol that serves this purposed is called SPV (see: SPV). Neutrino differs from SPV in that...".

> **output**. Output, transaction output, or TxOUt is an output...

Output is an output?.. Suggestion: "Output is an element of a Bitcoin transaction that defines the conditions under which the coins can be spent. Each transaction refers to existing outputs as inputs (see: input) and consumes them by providing the script that...".

**What editing does this need?**
I suggest taking every first sentence in each entry, strip it down to the basics, and make sure it really defines the term. Also, each definition should be as self-contained as possible. A reader should be able to understand the basic meaning of the term without referring much to other definitions.

ImranLorgat:
Fixes #462 

Used part of @s-tikhomirov 's proposed definition

### DIFF
--- a/glossary.asciidoc
+++ b/glossary.asciidoc
@@ -72,9 +72,10 @@ bitcoin, Bitcoin::
     Depending on context, could refer to the name of the currency unit (the coin), the network or the underlying enabling protocol. Written as bitcoin with a lowercase "b" usually refers to the currency unit. Bitcoin with an uppercase "B" usually refers to the protocol. See https://www.bitcoin.org for general information. The source code can be found at https://github.com/bitcoin/bitcoin.
 
 Bitcoin mining::
-    Bitcoin mining serves two purposes.
-    Firstly, Bitcoin transactions are validated and fixed in the sense that they cannot be double-spent.
-    Secondly, new bitcoin are being created within the coinbase of each block.
+    Bitcoin mining is the process of constructing a block from recent Bitcoin transactions and then solving a computational problem required as Proof of Work.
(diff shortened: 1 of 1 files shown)


ImranLorgat:
Duplicate of #482. For some reason got opened twice.

PR #493, opened 2020-09-18
node_operations: fees: BUG, wrong math

8go:
- "millionths per satoshi" ==> it is a rate, so it is just "millionths". Just like interest rate is 2% and not "2% per $" or "2%/$"
- rephrased and highlighted to make it more clear, more easy to follow
- changed 100,000 amount to 1,000,000 solely that at the end F_base and F_rate*P are different (different numbers make a better example)
- there was a BUG: in given example F_rate is 1 (not 100), correct total was 2 and not 101 as presented
- "F_rate = 10/1,000,000 * payment size" ==> wrong: ==> F_rate = 10/1,000,000 is correct
- put ( ) around F_rate * P to make it extra clear
- \Rightarrow is NOT showing correctly in "Preview" window, not sure why, \Rightarrow is correct Latex but not shown correctly here. I did not touch it. Maybe this is handled by post-processing.

### DIFF
--- a/node_operations.asciidoc
+++ b/node_operations.asciidoc
@@ -852,31 +852,31 @@ Running a Lightning node allows you to earn fees by routing payments across your
 
 Nodes compete for routing fees by setting their desired fee rate on each channel. Routing fees are set by two parameters on each channel: a fixed _base fee_ that is charged for any payment and an additional variable _fee rate_ that is proportional to the payment amount.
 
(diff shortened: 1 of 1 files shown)


rene78:
Ooops, I'm afraid there are some conflicting edits now. I was a little quicker than you https://github.com/lnbook/lnbook/pull/486.

8go:
Sorry, I had not looked at #486. Between all of us we will work it out :)

aantonop:
Please de-conflict @8go

8go:
@aantonop : Conflict resolved and I added some tiny-tiny additions to make it even more clear.
All of @rene78 changes were merged in. Thanks Rene. Liked the ppm notation.

PR #508, opened 2020-09-23
how ln works -typo

dangershony:


dangershony:
Sentence that was fixed `how many and well connected nodes your channels can reach`

PR #512, opened 2020-09-28
path-finding: source-based, mistakes, missing pieces

8go:
- feerate vs "fee rate"  : so far we have always used "fee rate" (2 words). I would stick with this for consistency. Also it is better English as feerate is an invented term.
- you wrote "FREERATE" --> ha ha very funny, a Freudian slip, but sorry, it will not be free. ;) smiling
- uppercasing
- commas
- avoid extremes and hype, avoid "very" : very expensive --> expensive
- successfull -> ...ful (one L)
- sentence simplifications
- Onion vs onion, lowercase or uppercase, since this is not a product name it should be lowercase
- BUG: the example does not contain the information needed, added "forward" field in onion
- BUG: 1% instead of 2% in text
- crucial part missing from a sentence
- etc

### DIFF
--- a/path-finding.asciidoc
+++ b/path-finding.asciidoc
@@ -69,22 +69,22 @@ For example one of the nodes becomes unreachable, doesn't have the channel balan
 Furthermore, there is no guarantee that the route chosen was the cheapest in terms of fees, or if a shorter path existed.
 As at the time of writing, this is a design trade-off made to protect user privacy.
 
-=== Paths are constructed from Destination to Source
-Let us assume our standard example in which Alice wants to send a payment of 100k satoshi on a path via Bob and Wei to Gloria.
-The Path obviously looks like (Alice)-->(Bob)-->(Wei)-->Gloria.
-However Bob and Wei will charge routing fees to forward the onion.
-As you already know nodes can charge two types of fees.
-First the base fee that will be charged for any successfull forwarding and settlement of and htlc.
-This fee is constant and does not depend on the amount that the node is supposed to forward.
-Additionally nodes might charge a freerate.
-The name rate alredy indicates that this fee depends on the amount that a node is supposed to forward.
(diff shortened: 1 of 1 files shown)


path-finding.asciidoc
```diff
@@ -93,13 +93,15 @@ She would find out that Bob would not forward the onion, if she believed that sh
          "channel": "357",
          "direction": 1,
          "satoshi": 103000,
+         "forward": 102000,
```
renepickhardt, line 96: not sure if that is correct. 
at the time I was writing this the API did not specify the forward amount in the onion. and also at other points we just have the amt_to_forward how I interpreted the `satoshi` field.

renepickhardt (commented):
Looks good besides the addition to the json of the onion. I think I see why you did this but I tried to be complient with the API output. 

Anyone any suggestions?

8go:
@renepickhardt 
The issue is that there is a contradiction unless we put the "forward" line in.

If the reader does not see the "forward" entry in the JSON, the reader will conclude that Bob
will just take his fee and forward the rest. Wei does the same, Wei takes his fees and passes on. 
Now Gloria receive too little money and will reject it. Unless each party knows what needs to be forwarded 
in some form or another, it is the destination/receiver who detects the missing funds. 

But your text states that BOB detects the lacking funds. So BOB must know how much to forward. 
And it was not visible in the JSON. So, I only see 2 options: a) a "forward" field is added to JSON or b) 
it is explained in the text that BOB computes the amount to be forwarded some way but that it is not shown 
in this sample JSON. I thought that option (a) is cleaner and easier to understand by the reader. 

If it is Bob who rejects the onion, he must know both values: received and to-be-forwarded. Somehow, the explanation needs to reflect that.

Let me know what you think and I rework if/as necessary.

renepickhardt:
I see the your thoughts and think they are very valid. In reality bob has announced his fees and knows what to forward. As we have discussed internally I will actually do some rewriting of the path finding chapter. so I will merge this for now and address this later if necessary.

PR #521, opened 2020-10-01
 path-finding: multipath: caution, exceptions apply

8go:
- added a cautionary note. 
- no need to add "for example 100k" as you are not referencing it again later
- likelyhood -> likelihood 
- etc


renepickhardt (approved):

renepickhardt:
actually very good catch noting that sometimes a smaller amount might not be forwarded by a node.

Issue #400, opened 2020-09-02
Topic suggestion: The Lightning Channel Graph: topology, attacks and privacy


seresistvanandras:
Inspired by a [recent tweet](https://twitter.com/renepickhardt/status/1298918019159207942?s=20) from René (@renepickhardt), it seems it would be great if we could mop up the misconceptions about the privacy of Lightning. To that end, jointly with Carla (@carlaKC), we would be excited to put down a more general and comprehensive chapter on the table which extensively reviews the channel graph not only from a privacy point of view but also from a structural point of view. Payment privacy on LN, among many other things, is largely a topological property. Therefore, it makes sense to give some grounding to this chapter by including purely topological considerations and descriptions of the LN channel graph.

**Please let us know your thoughts and feel free to suggest other sources/potential topics for this chapter!**

** What topic should be added? **
Lightning, in essence, is a network of payment channels. In this chapter, we would like to elaborate more on the network nature of LN from a network scientific point of view. First, we describe LN’s structural properties such as connectivity, centralization etc. Afterwards, we aim to analyze the channel graph’s robustness against topological and other related attacks.  Finally, we consider privacy, an important inherently topological property of LN. 

**1. Topological properties of the channel graph**

- Structure: how does the LN channel graph look like? Even though the graph changes over time, certain properties essentially remain unchanged (e.g. scale-freeness). These defining structural properties have quite some implications for the nature of the graph. 

- Connectivity: how hard is it to route payments in this graph? The small world nature of the graph. Centrality, transitivity of the channel graph. (https://arxiv.org/pdf/1904.10253.pdf,https://arxiv.org/pdf/1901.04972.pdf)

**2. Attacks on the channel graph**

- Robustness: how robust is this graph against failures or targeted attacks?

- More on topological attacks: node isolation, partitioning attacks (https://arxiv.org/pdf/1904.10253.pdf)

- Congestion attacks (https://arxiv.org/pdf/2002.06564.pdf)

- Time-Dilation Attacks on the Lightning Network (https://arxiv.org/pdf/2006.01418.pdf)

- Flood & Loot: A Systemic Attack On The Lightning Network (https://arxiv.org/pdf/2006.08513.pdf)

**3. Payment Privacy in LN**

- LN’s offered privacy guarantees: anonymity set of a node is its degree. The tradeoff between privacy and tx costs. The tradeoff between privacy and the amount of the payment (https://arxiv.org/pdf/1911.09432.pdf).  

[...]

PR #524, opened 2020-10-10
WIP: Security and Privacy of the Lightning network

seresistvanandras:
This is a work in progress PR. We (@s-tikhomirov, @carlaKC and me) plan to complete this chapter by mid/late November. Any suggestions, comments or feedback are welcome!


carlaKC:
Related to #400

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,597 @@
...
+
+The question of accepting and rejecting channels is a philosophical one.
+What if we end up with a Lightning Network where new nodes cannot participate because they cannot open any channels?
+Our suggestion is not to set an exclusive list of "mega-hubs" from which you will accept channels, but rather to accept channels in a manner that suits your risk preference.
+
+Some potential strategies are:
+
+* No risk: do not accept any incoming channels;
+* Low risk: only accept channels from nodes that have been present in the graph for a longer period, and have some long-lived channels;
+* Medium risk: only accept channels from
```
s-tikhomirov: from whom?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+It has been shown many times that the power of governments over money is often abused.
+Bitcoin offers individuals the possibility of storing and transferring their wealth without restrictions.
+
+Lightning continues this mission.
+It aims at providing a better user experience with instant payments while not sacrificing censorship resistance.
+However, the LN operates under a different security model, which introduces novel security and privacy challenges.
+
+= Definitions of privacy =
+
+Privacy is a complex topic.
```
carlaKC, line 17: Great paragraph!

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+In cryptographic systems, they are often centered around the mathematical properties of the cryptographic primitives such as ciphers, signatures, and hash functions.
+The security assumptions of the Lightning Network are that the ECDSA signatures, SHA-256 hash function, and other cryptographic functions used in the protocol behave within their security definitions.
+For example, we assume that it is practically impossible to find a pre-image of a hash function.
+This allows the LN to rely on the HTLC mechanism for the atomicity of multi-hop payments: nobody except the final recipient can reveal the payment secret and resolve the HTLC.
+We also assume a degree of connectivity in the network, namely, that Lightning channels form a connected graph (therefore, it is possible to find a path from any sender to any receiver) and that network messages get propagated within certain timeouts.
+
+Here are some possible models of adversaries in the Lightning Network.
+A regular "honest-but-curious" forwarding node can observe payment amounts, the immediately preceding and following nodes, the graph of announced channels with their capacities.
```
carlaKC: nit: "payment amounts" is a bit misleading here IMO, because there's a difference between the amount paid and the amount forwarded through a specific hop along the route.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+A very well-connected node can do the same but to a larger extent.
+For example, consider the developers of a popular wallet who maintain a node that their users connect to by default.
+This node would be responsible for routing a large share of payments to and from the users of that wallet.
```
carlaKC, line 34: The thought introduced in the previous paragraph isn't really concluded here - routing a large share of payments ... therefore can observe more? I think it's also safe to mention the single private channel that most wallets have by default.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+What may be the goals of a Lightning attacker?
+Information security is often described in terms of three main properties: confidentiality, integrity, and availability.
+Confidentiality means that the information only gets to intended recipients.
+Integrity implies that the information does not get altered in transit.
+Availability ensures that users may expect the system to be functioning most of the time.
+
+The important properties of the Lightning Network are mostly centered around confidentiality and availability.
+Here are some of the properties we may want to ensure:
+
+* nobody except the sender and the receiver knows the payment amount
```
carlaKC, line 48: should these be capitalized?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+* nobody can link senders and receivers
+* nobody can block an honest user from sending and receiving payments
```
carlaKC, line 50: nit: if we are going to refer to CIA here, then mention HMAC/that people can't meddle with the integrity of payments/messages that nodes send

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Instead, privacy research looks at anonymity compared to the _anonymity set_.
+
+The anonymity set is a central notion in privacy research.
+It refers to the set of identities such that, from an attacker's viewpoint, a given action could correspond to any of them.
+Consider a real-life example.
+Imagine you meet a person on a city street.
+What is their anonymity set from your point of view?
+If you don't know them personally, and without any additional information, their anonymity set roughly equals the city's population, including travelers.
+If you additionally consider their appearance, you may be able to roughly estimate their age and exclude the city residents who are obviously older or younger than the person in question from the anonymity set.
+Furthermore, if you notice that the person walks into the office of Company X using an electronic badge, the anonymity set shrinks to the number of Company X's employees and visitors.
```
carlaKC, line 80: Awesome explanation! I think you can safely make up a company name (maybe StarBlocks is already mentioned in this book)?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+This assures that the attacker has no information whatsoever.
+However, the real network leaks information that allows an attacker to narrow down the search.
+The smaller the anonymity set, the higher the chance of successful deanonymization.
+
+
+=== Differences between LN and Bitcoin in terms of privacy
+
+Bitcoin was initially perceived as an anonymous currency.
+Indeed, Bitcoin addresses are not linked to real identities.
+However, all transactions are broadcast in cleartext and can be analyzed.
```
carlaKC, line 107: Would it be too technical to mention that base layer has had its own privacy issues with transaction broadcast patterns?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Multiple companies deanonymize users of Bitcoin and other cryptocurrencies.
+
+At first glance, Lightning provides better privacy than Bitcoin because Lightning payments are not broadcast to the whole network.
+While this improves the privacy baseline, other properties of the Lightning protocol may make anonymous payments more challenging.
+For instance, larger payments have fewer routing options.
```
carlaKC: Would you say this is a property of the protocol or of the current graph structure?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+This may allow an adversary who controls well-capitalized nodes to route most large payments and discover payment amounts and probably other details.
```
carlaKC, line 113: "probably other details" is quite vague?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+This is necessary to create a new HTLC at every hop.
+The availability of payment amounts to intermediary nodes does not present an immediate threat.
+However, an _honest-but-curious_ intermediary node may use it as a part of a larger attack.
+
+
+== Linking senders and receivers
+
+An attacker might be interested in learning the sender and/or the receiver of a payment to reveal certain economic relationships.
+This breach of privacy could harm censorship resistance, as an intermediary node could censor payments to or from certain receivers or senders.
+Ideally, linking senders to receivers should not be possible to peers other than the sender and the receiver.
```
carlaKC, line 167: Arguably, the ideal would be that the sender and receiver don't even know each other? 
✨  insert rendez-vous routing magic here ✨

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+First, let us consider the *off-path adversary*.
+In the first step of this attack scenario, a potent off-path adversary deduces the individual balances in each payment channel via probing (described in a subsequent section) and forms a network snapshot at time _t_.
+It then runs the attack sometime later at time _t'_ and uses the differences between the two snapshots to infer information about payments that took place by looking at paths that have changed.
+In the simplest case, if only a single payment occurred between time _t'_ and _t_, the adversary observes a single path where the balances have changed by the same amounts.
+Thus, the adversary learns everything about this payment: the sender, the recipient, and the amount.
+If multiple payment paths overlap, the adversary needs to apply heuristics to identify such overlap and separate the payments.
+
+Now, we turn our attention to an *on-path adversary*.
+Such an adversary might seem convoluted.
+However, the single most central node can already observe close to 50% of all LN payments, while the four most central nodes observe an average of 72% payments.
```
carlaKC: Flatten these into one sentence?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Bob decrypts the payment onion and realizes that the payment is intended for him.
+He looks up his local store of payment secrets, searches for the preimage that corresponds to the payment hash but does not find one.
+Following the protocol, Bob returns the "unknown payment hash" error to Alice, who relays it back to Mallory.
+As a result, Mallory knows that the payment _could have succeeded_ if the payment hash was real.
+Therefore, Mallory can update her estimation of Alice's balance from "between zero and 1 million" to "between 500 thousand and one million."
+Another scenario happens if Alice's balance is lower than 500 thousand satoshis.
+In that case, Alice is unable to forward the payment and returns the "insufficient balance" error to Mallory.
+Mallory updates her estimation from "between zero and 1 million" to "between zero and 500 thousand."
+
+Note that in any case, Mallory's estimation becomes twice as precise after just one probing!
```
carlaKC, line 239: This is a great paragraph 🕺

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+The probing attack can be further used to link senders and receivers, as described in the previous section.
+
+At this point, you may ask: why does the Lightning Network do such a poor job at protecting its users' private data?
+Wouldn't it be better to not reveal to the sender why and where the payment has failed?
+Indeed, this could be a potential countermeasure, but it has significant drawbacks.
+Lightning has to strike a careful balance between privacy and efficiency.
+Remember that regular nodes don't know balance distributions in remote channels.
+Therefore, payments can (and often do) fail because of insufficient balance at an intermediary hop.
+Error messages allow the sender to exclude the failing channel from consideration when constructing another route.
+A popular Lightning wallet Zap even performs probing internally to check whether a constructed route can really handle a payment.
```
carlaKC, line 269: Safe to say that many applications do this, it's not just zap

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+==== Channel liquidity lockup
+
+A channel liquidity lockup attack is comparable to a channel jamming attack in that it routes payments through a channel and holds them so that the channel is unusable.
+Rather than locking up slots on the channel commitment, this attack routes large HTLCs through a target channel that consumes all the channel's available bandwidth.
+This attack's capital commitment is higher than the commitment jamming attack because the attacking node needs more funds to route failed payments through the target.
+
+References:
+
+* Mizrahi, A., Zohar, A. https://arxiv.org/abs/2002.06564[Congestion Attacks in Payment Channel Networks]
```
carlaKC, line 343: I think there's another paper I need to link here, anybody remember one specifically about liquidity lockup?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Specifically, one clusters LN nodes into a single address if their aliases are similar with respect to some string similarity metric.
+
+Another method to cluster LN nodes is applying their IP or Tor addresses.
+If the same IP or Tor addresses correspond to different LN nodes, these nodes are likely controlled by the same user.
+
+_Countermeasures_: For more privacy, aliases should be sufficiently different from one another.
+While the public announcement of IP addresses may be unavoidable for those nodes that wish to have incoming channels in the LN, linkability across nodes of the same user can be mitigated if the clients for each node are hosted with different service providers and thus IP addresses.
+
+*Cross-Layer Linking: Lightning Nodes and Bitcoin Entities*
+Associating LN nodes to Bitcoin entities is a serious breach of privacy that is exacerbated by the fact that most LN nodes publicly expose their IP addresses.
```
carlaKC, line 411: Isn't the biggest threat here eclipse attacks on the bitcoin node?

carlaKC (commented):
Nice to read all of this coming together!

seresistvanandras:
Dear authors: @renepickhardt, @Roasbeef and @aantonop! We are ready with our first feature-full draft. We would greatly appreciate if you could have a review of this chapter. Please let us know your thoughts, comments, critique! :)

aantonop:
We are reviewing! Thank you @seresistvanandras and others

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+In this chapter, we look at the issues related to the security and privacy of the LN.
+
+= Why is privacy important? =
+
+The key value proposition of cryptocurrency is censorship resistant money.
+It has been shown many times that the power of governments over money is often abused.
+Bitcoin offers individuals the possibility of storing and transferring their wealth without restrictions.
+
+Lightning continues this mission.
+It aims at providing a better user experience with instant payments while not sacrificing censorship resistance.
```
Roasbeef: I think this sentence is a bit too strong to present without first attempting to define what is meant by "censorship resistance" here. As it's a layer two system, it makes certain tradeoffs at a fundamental level in favor of faster lower cost payments compared to the based layer. As a result, it makes addition assumptions which may be characterized as it having less "censorship resistance" compared to the base layer.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+Privacy is a complex topic.
+It is hard to precisely define what we mean by privacy.
+Therefore, the question: "Is Lightning private?" has no direct answer.
+Privacy researchers approach this issue as follows.
+First, they define a _security model_ that specifies what an adversary is capable of and aims to achieve.
+Then, they describe the relevant properties of the system and check whether it conforms to the requirements.
+A security model is based on a set of underlying _security assumptions_.
+In cryptographic systems, they are often centered around the mathematical properties of the cryptographic primitives such as ciphers, signatures, and hash functions.
+The security assumptions of the Lightning Network are that the ECDSA signatures, SHA-256 hash function, and other cryptographic functions used in the protocol behave within their security definitions.
```
Roasbeef, line 25: Only an intro paragraph, but I think one thing worth mentioning here is that since LN is a _networked_ protocol, any security model must also factor in how interaction with the network and also topological constraints affect the privacy of the network. As an example, since we use an onion routing protocol for HTLC extension, we by default don't assume "protection" against a global passive adversary, as such an adversary can use packet+timing information to deduce payment flows.

security_privacy_ln.asciidoc
Roasbeef: missing an "and" before the last "the" here?

Roasbeef: Worth mentioning here that with MPP payment amounts are slightly more obfuscated given non uniform split sizes, and also that tech like multi-hop locks also removes the payment ID correlation as well.

security_privacy_ln.asciidoc
Roasbeef, line 34: I think there's an implicit/missing conclusion here? Seems y'all intend to convey that this wallet author learns information about the payments/patterns of each user that they serve?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+We also assume a degree of connectivity in the network, namely, that Lightning channels form a connected graph (therefore, it is possible to find a path from any sender to any receiver) and that network messages get propagated within certain timeouts.
+
+Here are some possible models of adversaries in the Lightning Network.
+A regular "honest-but-curious" forwarding node can observe payment amounts, the immediately preceding and following nodes, the graph of announced channels with their capacities.
+A very well-connected node can do the same but to a larger extent.
+For example, consider the developers of a popular wallet who maintain a node that their users connect to by default.
+This node would be responsible for routing a large share of payments to and from the users of that wallet.
+An even stronger security model implies that multiple nodes are under adversarial control.
+If it so happens that two colluding nodes are on the same payment path, they would understand that they are forwarding HTLCs belonging to the same payment, as such HTLCs have the same payment hash.
```
Roasbeef, line 36: See comment re payment amts and multi-hop locks above.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+What may be the goals of a Lightning attacker?
+Information security is often described in terms of three main properties: confidentiality, integrity, and availability.
+Confidentiality means that the information only gets to intended recipients.
+Integrity implies that the information does not get altered in transit.
+Availability ensures that users may expect the system to be functioning most of the time.
+
+The important properties of the Lightning Network are mostly centered around confidentiality and availability.
+Here are some of the properties we may want to ensure:
```
Roasbeef, line 46: Who's "we" here, it reads as if the authors are attempting to retroactively apply their own security models to LN, rather than attempt to analyze the details of the system and the resulting emerging security attributes of it.

security_privacy_ln.asciidoc
Roasbeef, line 50: Super strong statement, and I think just given the availability requirements of channels, this only applies to _direct_ channels in the system (Alice and Bob are connected with no one in between them).

security_privacy_ln.asciidoc
Roasbeef, line 48: These statements all feel rather strong, though I think the context here is that these are just _hypothetical_ properties we may want to have in an ideal payment system?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+* nobody except the sender and the receiver knows the payment amount
+* nobody can link senders and receivers
+* nobody can block an honest user from sending and receiving payments
+
+For each privacy goal and security model, there is a certain probability that an attacker succeeds.
+This probability depends on various factors, such as the size and the structure of the network.
+Other things equal, it is generally easier to attack a small network rather than a large one.
```
Roasbeef, line 54: All things equal?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Each payment may be assigned a sender anonymity set and a receiver anonymity set.
+Ideally, the anonymity set consists of all the users of the network.
+This assures that the attacker has no information whatsoever.
+However, the real network leaks information that allows an attacker to narrow down the search.
+The smaller the anonymity set, the higher the chance of successful deanonymization.
+
+
+=== Differences between LN and Bitcoin in terms of privacy
+
+Bitcoin was initially perceived as an anonymous currency.
```
Roasbeef: Was it? Very strong statement to make here, Bitcoin has no explicit anonymity enhancing properties.

security_privacy_ln.asciidoc
Roasbeef: may have* 

As it depends on the topology of the graph itself and also the make up of the channels connecting them.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Multiple companies deanonymize users of Bitcoin and other cryptocurrencies.
+
+At first glance, Lightning provides better privacy than Bitcoin because Lightning payments are not broadcast to the whole network.
+While this improves the privacy baseline, other properties of the Lightning protocol may make anonymous payments more challenging.
+For instance, larger payments have fewer routing options.
+This may allow an adversary who controls well-capitalized nodes to route most large payments and discover payment amounts and probably other details.
+Another relevant difference between Lightning and Bitcoin is that Lightning nodes maintain a permanent identity, whereas Bitcoin nodes do not.
+A Bitcoin user can easily switch nodes used to receive blockchain data and broadcast transactions.
+A Lightning user, on the contrary, sends and receives payments through the nodes they have opened channels with.
+Moreover, the Lightning protocol assumes that routing nodes announce their IP address in addition to their node ID.
```
Roasbeef, line 117: Worth mentioning that Tor is available, so IP addrs are not required.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+This creates a permanent link between node IDs and IP addresses, which may be dangerous, considering that an IP address is often an intermediary step in anonymity attacks linked to the user's physical location and, in most cases, real-world identity.
+It is possible to use Lightning over Tor, but many nodes do not use this functionality.
```
Roasbeef: I think this statement needs to be cited, from my PoV I see Tor usage rising by the week for new nodes.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+A Lightning user, when sending a payment, has its neighbors in its anonymity set.
+Specifically, a routing node only knows the immediately preceding and following nodes.
+The routing node does not know whether its immediate neighbors in the payment route are the ultimate sender or receiver.
+Therefore, the anonymity set of a node in Lightning roughly equals its neighbors.
+Similar logic applies to payment receivers.
+Many users open only a handful of payment channels, therefore, limiting their anonymity sets.
+Moreover, in Lightning, the anonymity set is static or at least slowly changing.
+In contrast, one can achieve significantly larger anonymity sets in on-chain CoinJoin transactions.
```
Roasbeef, line 128: Seems like a bit of an apples-and-oranges comparison: the off-chain layer can employ techniques such a probing and circular payments to self to insert noise into its payment pattern which allows it to passively increase maintain its anonymity set compared to techniques employed using the persistent on-chain transaction graph.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+CoinJoin transactions with anonymity sets larger than 50 are quite frequent.
+Typically, the anonymity sets in a CoinJoin transaction correspond to a dynamically changing set of users.
+
+[[anonymity-set]]
+.The anonymity set of Alice and Bob constitutes their neighbors.
+image:images/anon-set.png[]
+
+Finally, Lightning users can also be denied service, having their channels blocked by the attacker.
```
Roasbeef: Unclear at which level this "blocking" would occur: on-chain, refuse of update, HTLC jamming, etc?

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Forwarding payments requires capital - a scarce resource! - to be temporarily blocked in in-flight HTLCs along the route.
+An attacker may send many payments but fail to finalize them, occupying honest user's capital for long periods.
+This attack vector is not present (or at least not as obvious) in Bitcoin.
```
Roasbeef, line 139: An analogous situation would be a user being unable to transact due to full blocks and/or prohibitive fees.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+In summary, while some aspects of the Lightning Network's architecture suggest that it is a step forward in terms of privacy compared to Bitcoin, other properties of the protocol may make attacks on privacy easier.
+Thorough research is needed to evaluate what privacy guarantees the LN provides and improve the state of affairs.
+We are happy to report that multiple researchers teams are working on Lightning privacy.
+In part, this chapter summarizes the research papers available at the time of writing in late 2020.
```
Roasbeef, line 144: I think this section is missing the clear (imo) advantage privacy wise: you as a user _select_ the set of nodes you wish to interact with in order to send payments, and the transcript of your payments (the partial information) is known only to those nodes. When transacting on-chain, the transaction graph is made public for all of time, and you don't necessarily select which nodes hear about your transactions, or which miners mined the transactions. 

Consider the following statement: so far in 2020, we know _exactly_ how many Bitcoin transactions have happend off-chain. However, no one is able to make a strong claim w.r.t the number of LN transactions that have happened off-chain.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+= Attacks on Lightning =
+
+Recent research describes various ways in which the security and privacy of the LN may be compromised.
+
+== Observing payment amounts
+
+One of the goals for a privacy-preserving payment system is to hide the payment amount from uninvolved parties.
+The Lightning Network is an improvement over layer-one in this regard.
+While Bitcoin transactions are broadcast in cleartext and can be observed by anyone, Lightning payments only travel through a few nodes along the payment path.
+However, intermediary nodes do see the payment amount.
```
Roasbeef: Worth mentioning here that due to MPP, the payment amount they see may not necessarily be the actual total payment amount.

security_privacy_ln.asciidoc
Roasbeef: Yep, such techniques can allow us to achieve sender-receiver privacy (sender doesn't know the true node ID of the receiver, the reverse already exists as is).

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Ideally, linking senders to receivers should not be possible to peers other than the sender and the receiver.
+In the following, we will consider two types of adversaries: the off-path and the on-path adversary.
+An off-path adversary tries to assess the sender and the receiver of a payment without participating in the payment routing process.
+An on-path adversary can leverage any information it might gain by routing the payment of interest.
+
+First, let us consider the *off-path adversary*.
+In the first step of this attack scenario, a potent off-path adversary deduces the individual balances in each payment channel via probing (described in a subsequent section) and forms a network snapshot at time _t_.
+It then runs the attack sometime later at time _t'_ and uses the differences between the two snapshots to infer information about payments that took place by looking at paths that have changed.
+In the simplest case, if only a single payment occurred between time _t'_ and _t_, the adversary observes a single path where the balances have changed by the same amounts.
+Thus, the adversary learns everything about this payment: the sender, the recipient, and the amount.
```
Roasbeef, line 176: Though it's a simple model, I think this should also mention the existence of unadvertised channels as well, and the potential that a "shadow" network exists beyond the graph view of this attacker.

security_privacy_ln.asciidoc
Roasbeef: Citation? Or at least elaboration.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+The probing attack proceeds as follows.
+Say, the attacker Mallory wants to reveal Alice's balance of a public channel between Alice and Bob.
+Suppose the total capacity of that channel is 1 million satoshis.
+Alice's balance could be anything from zero to 1 million satoshis (to be precise, the estimate is a bit tighter due to channel reserve, but we don't account for it here for simplicity).
+Mallory opens a channel with Alice with 1 million satoshis and sends 500 thousand satoshis to Bob via Alice using a _random number_ as the payment hash.
+Of course, this number does not correspond to any known payment secret.
+Therefore, the payment will fail.
+The question is: how exactly will it fail?
+
+There are two scenarios.
```
Roasbeef, line 228: Simplified example, but this is missing a bit of nuance re the existence of the channel level flow control constraints that may be slightly out of sync with the advertised routing constraints. As an example, let's say a channel only allows 5 HTLCs for w/e reason in this case when a prober attempts to attach a 6th HTLC, they'll get a "temp chan failure" even though the channel may have had sufficient bandwidth to carry the channel. 

Another value to consider is the `max_htlc` setting, which may not allow a prober to deduce the full available bandwidth amount in the channel in a single go.

Roasbeef (changes requested):
Chunking my review a bit to avoid a massive dump of comments, will likely have 2/3 more review chunks to complete my final pass. 

My initial impression though is that this is an amazing document, and I don't think there exists another comparable approachable single piece of writing that attempts to approach security+privacy of LN from a similarly objective angle. Amazing contribution!

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+image:images/anon-set.png[]
+
+Finally, Lightning users can also be denied service, having their channels blocked by the attacker.
+Forwarding payments requires capital - a scarce resource! - to be temporarily blocked in in-flight HTLCs along the route.
+An attacker may send many payments but fail to finalize them, occupying honest user's capital for long periods.
+This attack vector is not present (or at least not as obvious) in Bitcoin.
+
+In summary, while some aspects of the Lightning Network's architecture suggest that it is a step forward in terms of privacy compared to Bitcoin, other properties of the protocol may make attacks on privacy easier.
+Thorough research is needed to evaluate what privacy guarantees the LN provides and improve the state of affairs.
+We are happy to report that multiple researchers teams are working on Lightning privacy.
```
Trxlz:
```suggestion
We are happy to report that multiple research teams are working on Lightning privacy.
```

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Indeed, this could be a potential countermeasure, but it has significant drawbacks.
+Lightning has to strike a careful balance between privacy and efficiency.
+Remember that regular nodes don't know balance distributions in remote channels.
+Therefore, payments can (and often do) fail because of insufficient balance at an intermediary hop.
+Error messages allow the sender to exclude the failing channel from consideration when constructing another route.
+A popular Lightning wallet Zap even performs probing internally to check whether a constructed route can really handle a payment.
+
+There are other potential countermeasures against channel probing.
+First, it is hard for an attacker to target unannounced channels.
+Second, nodes that implement JIT routing are less prone to the attack.
```
Roasbeef: I think that should be relaxed to "may be" less prone to probing, as to my knowledge it hasn't really been deployed at a large scale to be able to draw solid empirical conclusions. It seems as though JIT routing would also exacerbate the issue of HTLC slot utilization, given that a single prob across `N` nodes would then cause each of those nodes to attempt to generate up to `K` routes themselves to attempt to rebalancing things, each of which may actually fail.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+This measure largely ensured that the transactions that consume network resources will eventually pay their chain fees.
+The minimum relay fee is acceptable to regular users but would hurt attackers financially if they tried to spam the network.
+While some transactions may not make it into valid blocks during high fee environments, these measures have largely been effective at disincentivizing this type of spam.
+
+=== DoS in Lightning
+
+Similarly to Bitcoin, the Lightning Network charges fees for the use of its public resources, but in this case, the resources are public channels, and the fees come in the form of routing fees.
+The ability to route payments through nodes in exchange for fees provides the network with a large scalability benefit - nodes that are not directly connected can still transact - but it comes at the cost of exposing a public resource that must be protected against DoS attacks.
+
+When a Lightning node forwards a payment on your behalf, it uses bandwidth to update its commitment transaction, and the amount of the payment is reserved in their channel balance until it is settled or failed.
```
Roasbeef: Unclear if "bandwidth" here refers to payment bandwidth, or actual data bandwidth, works in either interpretation though IMO.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+This is great for legitimate users, who wouldn't like to pay for failed attempts, but also allows attackers to costlessly consume nodes' resources - much like the low-fee transactions on Bitcoin that never end up paying miner fees.
+
+At the time of writing, a discussion is https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-June/002734.html[ongoing] on the lightning-dev mailing list as to how best address this issue.
+
+=== Known DoS attacks
+
+There are two known DoS attacks on public Lightning channels which render a target channel, or a set of target channels, unusable.
+Both attacks involve routing payments through a public channel, then holding them until their timeout, thus maximizing the attack's duration.
+The requirement to fail payments to not pay fees is fairly simple to meet because malicious nodes can simply reroute payments to themselves.
+In the absence of fees for failed payments, the only cost to the attacker is the on-chain cost of opening a channel to dispatch these payments through, which can be trivial in low fee environments.
```
Roasbeef, line 324: The cost also includes worst case closing 2x the channels (if they're doing it via circular routes), and also the opportunity cost of the Bitcoin they need to put into channels to launch the attack.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+==== Commitment jamming
+
+Lightning nodes update their shared state using asymmetric commitment transactions, on which HTLCs are added and removed to facilitate payments.
+Each party is limited to a total of https://github.com/lightningnetwork/lightning-rfc/blob/c053ce7afb4cbf88615877a0d5fc7b8dbe2b9ba0/02-peer-protocol.md#the-open_channel-message[483] on the commitment at a time.
+A channel jamming attack allows an attacker to render a channel unusable by routing 483 payments through the target channel and holding them until they timeout.
+
+It should be noted that this limit is arbitrary and was chosen in the specification to ensure that all the HTLCs can be swept in a https://github.com/lightningnetwork/lightning-rfc/blob/master/05-onchain.md#penalty-transaction-weight-calculation[single justice transaction].
```
Roasbeef: It isn't arbitrary in that it was selected to respect the _policy_ limits on the max size of a transaction in the block.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+While this limit _may_ be increased, transactions are still limited by the block size, so the number of slots available is likely to remain limited.
```
Roasbeef, line 333: One missing component here is that it's possible to essentially lift this limit (by using indirect commitments), instead letting nodes manage the free HTLC slots as if they were space in a queue. Modern implementations also feature something commonly referred to as an "HTLC interceptor" that allows them to make accept/deny decisions based on the current channel state. In a sense, the problem is similar to TCP congestion control/avoidance and buffer bloat itself.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+However, users tend to use similar aliases for their own different nodes.
+For example, _LNBig.com Billing_ is likely owned by the same user as the node with alias _LNBig.com_.
+Given this observation, one can cluster LN nodes applying their node aliases.
+Specifically, one clusters LN nodes into a single address if their aliases are similar with respect to some string similarity metric.
+
+Another method to cluster LN nodes is applying their IP or Tor addresses.
+If the same IP or Tor addresses correspond to different LN nodes, these nodes are likely controlled by the same user.
+
+_Countermeasures_: For more privacy, aliases should be sufficiently different from one another.
+While the public announcement of IP addresses may be unavoidable for those nodes that wish to have incoming channels in the LN, linkability across nodes of the same user can be mitigated if the clients for each node are hosted with different service providers and thus IP addresses.
```
Roasbeef, line 408: Tor also goes a long way here given that onion services addresses are effectively infinite, while IPv4 adresses are still scarce.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+For instance, it can DDoS them or make them non-operational by any means.
+It turns out that if the adversary chooses nodes randomly, then scale-free networks like the LN are robust against node-removal attacks.
+This is because a random node lies on the periphery with a small number of connections, therefore playing a negligible role in the network's connectivity.
+However, if the adversary is more prudent, it can target the most well-connected nodes.
+Not surprisingly, the LN and other scale-free networks are _not_ robust against targeted node-removal attacks.
+
+On the other hand, the adversary could be more stealthy.
+Several topology-based attacks target a single node or a single payment channel.
+For example, an adversary might be interested in exhausting a certain payment channel's capacity on purpose.
+More generally, an adversary can deplete all the outgoing capacity of a node to knock it down from the routing market.
```
Roasbeef, line 486: How so? Outgoing capacity is only "shifted around" and never really depleted in a sense. Also users are able to disable certain chanenls to protect the outgoing capacity of those channels.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+It would be beneficial to explore the temporal nature of the LN graph in more depth.
+
+
+=== Economic incentives and graph structure
+The Lightning graph forms spontaneously.
+Nodes connect to each other based on mutual interest.
+As a result, incentives drive graph development.
+Let's describe some of the relevant incentives:
+
+  * Rational incentives.
```
Roasbeef, line 547: A new incentive comes in the form of Lightning Pool: aside from the _potential_ to earn routing fees, nodes are also able to earn passive yield simply by allocating capital to the network where it's most demanded.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Put differently, a profit-seeker node will try to locate itself in the _center_ of the channel graph or close to it.
+Given these incentives and the fact that capital is also unevenly distributed, it is unlikely that LN will evolve into a mesh-like network.
+
+= Practical advice for users to protect their privacy =
+
+We're still in the early stages of the Lightning Network.
+Many of the concerns listed in this chapter are likely to be addressed as it matures and grows.
+In the meantime, there are some measures that you can take to guard your node against malicious users; something as simple as updating the default parameters that your node runs with can go a long way in hardening your node.
+
+== Private channels
```
Roasbeef, line 568: Another benefit is that one is able to more effectively guard/utilize its channel bandwidth: given that these private channels will only be used to receiving or sending directly to that node.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+If you primarily intend to use your node for personal sends and receives, there is little need to open public channels to the network.
+Since you are not exposing your public channels to the network, you eliminate the risk of a denial-of-service attack on your node.
+
+== Routing considerations
+
+As covered in the <<Denial of service>> section, nodes that open public channels expose themselves to the risk of a series of attacks on their channels.
+While mitigations are being developed on the protocol level, there are many steps that a node can take to protect against denial of service attacks on their public channels.
+
+* Minimum HTLC size: on channel open, your node can set the minimum HTLC size that it will accept.
+Setting a higher value ensures that each of your available channel slots cannot be occupied by a very small payment.
```
Roasbeef, line 579: Another related setting here is `max_htlc`, which can be used to help mitigate probing attacks somewhat, and also let a node control the exposure it'll have to a given payment flow.

security_privacy_ln.asciidoc
nopara73, line 36: "if it so" is strange

```suggestion
If two colluding nodes are on the same payment path, they would understand that they are forwarding HTLCs belonging to the same payment, as such HTLCs have the same payment hash.
```

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Alice's balance could be anything from zero to 1 million satoshis (to be precise, the estimate is a bit tighter due to channel reserve, but we don't account for it here for simplicity).
+Mallory opens a channel with Alice with 1 million satoshis and sends 500 thousand satoshis to Bob via Alice using a _random number_ as the payment hash.
+Of course, this number does not correspond to any known payment secret.
+Therefore, the payment will fail.
+The question is: how exactly will it fail?
+
+There are two scenarios.
+If Alice has more than 500 thousand satoshis on her side of the channel to Bob, she forwards the payment.
+Bob decrypts the payment onion and realizes that the payment is intended for him.
+He looks up his local store of payment secrets, searches for the preimage that corresponds to the payment hash but does not find one.
```
nopara73, line 231: Use `pre-image` instead of `preimage` for consistency as the first version was used previously.

```suggestion
He looks up his local store of payment secrets, searches for the pre-image that corresponds to the payment hash but does not find one.
```

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+The ability to route payments through nodes in exchange for fees provides the network with a large scalability benefit - nodes that are not directly connected can still transact - but it comes at the cost of exposing a public resource that must be protected against DoS attacks.
+
+When a Lightning node forwards a payment on your behalf, it uses bandwidth to update its commitment transaction, and the amount of the payment is reserved in their channel balance until it is settled or failed.
+In successful payments, this is acceptable because the node is eventually paid out its fees.
+Failed payments do not incur fees in the current protocol.
+This allows nodes to costlessly route failed payments through any channels.
+This is great for legitimate users, who wouldn't like to pay for failed attempts, but also allows attackers to costlessly consume nodes' resources - much like the low-fee transactions on Bitcoin that never end up paying miner fees.
+
+At the time of writing, a discussion is https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-June/002734.html[ongoing] on the lightning-dev mailing list as to how best address this issue.
```
nopara73, line 317: Specifying timing twice is redundant. Also added a `to`.

```suggestion
A discussion is https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-June/002734.html[ongoing] on the lightning-dev mailing list as to how best to address this issue.
```

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+* Shadow channels: nodes that wish to open large channels to a single target can instead open a single public channel to the target and support it with further private channels called [shadow channels](joost on TFTC).
+These channels can still be used for routing but are not announced to potential attackers.
+
+References:
+
+* Jager, J. https://anchor.fm/tales-from-the-crypt/episodes/197-Joost-Jager-ekghn6[Tales from the Crypt Episode 197]
+
+== Accepting channels
+At present, Lightning nodes struggle with bootstrapping inbound liquidity.
+This issue has lead to an attitude of accepting any channel that another peer attempts to open to your node.
```
nopara73: tense

```suggestion
This issue has led to an attitude of accepting any channel that another peer attempts to open to your node.
```

security_privacy_ln.asciidoc
nopara73:
```suggestion
Unlike trivial scaling solutions: custodial Bitcoin banks, LN aims to scale Bitcoin without compromising on self-custody, which leads to greater censorship resistance in the Bitcoin ecosystem.
```

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+An even stronger security model implies that multiple nodes are under adversarial control.
+If it so happens that two colluding nodes are on the same payment path, they would understand that they are forwarding HTLCs belonging to the same payment, as such HTLCs have the same payment hash.
+
+What may be the goals of a Lightning attacker?
+Information security is often described in terms of three main properties: confidentiality, integrity, and availability.
+Confidentiality means that the information only gets to intended recipients.
+Integrity implies that the information does not get altered in transit.
+Availability ensures that users may expect the system to be functioning most of the time.
+
+The important properties of the Lightning Network are mostly centered around confidentiality and availability.
```
nopara73, line 45: Why? Is is because integrity comes out of the box? If so, it's worth mentioning.

security_privacy_ln.asciidoc
nopara73: My reading was it's the aim of a system and not the properties of the discussed system.

security_privacy_ln.asciidoc
nopara73, line 50: This third one belongs to availability, while the first two to confidentiality, so I think these should not be in the same list.

security_privacy_ln.asciidoc
nopara73: I think not, that changes the meaning. But I agree that it's strange phrasing.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+=== Anonymity set
+
+What does it mean to deanonymize someone?
+In simple terms, deanonymization implies linking some action with a person's "real world" identity, such as their name or physical address.
+In privacy research, the notion of deanonymization is more nuanced.
+First, we are not necessarily talking about names and addresses.
+Discovering someone's IP address or telephone number may also be considered deanonymization.
+A piece of information that allows linking a user's action to their previous actions is referred to as _identity_.
+Second, deanonymization is not binary.
+A user is not either fully anonymous or completely deanonymized.
```
nopara73:
```suggestion
A user is not fully anonymous nor completely deanonymized.
```

security_privacy_ln.asciidoc
nopara73: ACK, even the comments under the first announcement on the cypherpunk mailing list were concerned about the weak privacy of this system.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+However, the real network leaks information that allows an attacker to narrow down the search.
+The smaller the anonymity set, the higher the chance of successful deanonymization.
+
+
+=== Differences between LN and Bitcoin in terms of privacy
+
+Bitcoin was initially perceived as an anonymous currency.
+Indeed, Bitcoin addresses are not linked to real identities.
+However, all transactions are broadcast in cleartext and can be analyzed.
+Multiple companies deanonymize users of Bitcoin and other cryptocurrencies.
```
nopara73:
```suggestion
Multiple companies were established to deanonymize users of Bitcoin and other cryptocurrencies.
```

security_privacy_ln.asciidoc
nopara73: I think that's just a random detail there as there are many off-chain privacy leaks those are equally as important or even more important, like sending xpubs to servers.

Maybe what should be done here rather is to make clear the difference between the inherent privacy issues of transactions being public and the additional privacy issues that the Bitcoin ecosystem introduced on top of that.

security_privacy_ln.asciidoc
nopara73: @carlaKC You suggested in https://github.com/lnbook/lnbook/pull/524/files#r525399837:

> Would it be too technical to mention that base layer has had its own privacy issues with transaction broadcast patterns?

For that I may ask: would you say this is a property of the protocol or of the current usage patterns?

IMO my solution: "differentiating between inherent protocol privacy issues and additional privacy issues" should be also applied here: https://github.com/lnbook/lnbook/pull/524/files#r541590406

The thing is, discussing LN and on-chain privacy in isolation is fine for an academic paper as the models are hard to untangle with all the variables included, however a book should attempt to take a holistic approach and consider all the variables.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+
+At first glance, Lightning provides better privacy than Bitcoin because Lightning payments are not broadcast to the whole network.
+While this improves the privacy baseline, other properties of the Lightning protocol may make anonymous payments more challenging.
+For instance, larger payments have fewer routing options.
+This may allow an adversary who controls well-capitalized nodes to route most large payments and discover payment amounts and probably other details.
+Another relevant difference between Lightning and Bitcoin is that Lightning nodes maintain a permanent identity, whereas Bitcoin nodes do not.
```
nopara73, line 114: Although Bitcoin nodes do not assume a permanent identity, wallet clustering achieves similar outcomes for wallets without/under utilized/or ineffective privacy techniques.

security_privacy_ln.asciidoc
nopara73: According to https://1ml.com/statistics: 

- Number of Nodes: 15,079
- Tor Onion Service Nodes: 2,715

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+A Bitcoin user can easily switch nodes used to receive blockchain data and broadcast transactions.
+A Lightning user, on the contrary, sends and receives payments through the nodes they have opened channels with.
+Moreover, the Lightning protocol assumes that routing nodes announce their IP address in addition to their node ID.
+This creates a permanent link between node IDs and IP addresses, which may be dangerous, considering that an IP address is often an intermediary step in anonymity attacks linked to the user's physical location and, in most cases, real-world identity.
+It is possible to use Lightning over Tor, but many nodes do not use this functionality.
+
+A Lightning user, when sending a payment, has its neighbors in its anonymity set.
+Specifically, a routing node only knows the immediately preceding and following nodes.
+The routing node does not know whether its immediate neighbors in the payment route are the ultimate sender or receiver.
+Therefore, the anonymity set of a node in Lightning roughly equals its neighbors.
```
nopara73, line 124: Not sure about this statement. Researchers are trying to fit the onion routing into their anonymity set (entropist) models with little success: https://www.freehaven.net/anonbib/cache/entropist.pdf

> Starting from an entropist conception of anonymity has led to a focus on system properties that are not the most important to system security, has led to system assumptions that are not reasonable in practice, has led to adversary models that are not reasonable in practice, and has led to both theory and system design for anonymous communication that fail in significant ways.

security_privacy_ln.asciidoc
nopara73: Yup, I think my criticism is aligned with that: https://github.com/lnbook/lnbook/pull/524/files#r541603355

Coinjoins are more fit for an entropist analysis than onion routing.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+An off-path adversary tries to assess the sender and the receiver of a payment without participating in the payment routing process.
+An on-path adversary can leverage any information it might gain by routing the payment of interest.
+
+First, let us consider the *off-path adversary*.
+In the first step of this attack scenario, a potent off-path adversary deduces the individual balances in each payment channel via probing (described in a subsequent section) and forms a network snapshot at time _t_.
+It then runs the attack sometime later at time _t'_ and uses the differences between the two snapshots to infer information about payments that took place by looking at paths that have changed.
+In the simplest case, if only a single payment occurred between time _t'_ and _t_, the adversary observes a single path where the balances have changed by the same amounts.
+Thus, the adversary learns everything about this payment: the sender, the recipient, and the amount.
+If multiple payment paths overlap, the adversary needs to apply heuristics to identify such overlap and separate the payments.
```
nopara73, line 177: Since my LN knowledge is quite lacking, I can't claim to fully understand what's going on here, but I created some illustration to make it easier for the reader to grasp the basic concept at least. If you want to add it or create something similar incorporating the amount attack itself that could be helpful to improve my understanding without reading the rest of the book.
image

security_privacy_ln.asciidoc
nopara73: Specify when were these statistics are collected with citation as @Roasbeef suggested: https://github.com/lnbook/lnbook/pull/524/files#r537747008

Also if this information could be estimated from an already existing public website, then it could be useful to document how so in the future, the numbers could be easily updated.

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+We note that this leakage could be minimized with multi-part payments or with large capacity payment channels.
+Similarly, payment channels with small time-lock deltas could be excluded from a payment path.
+More precisely, a payment channel cannot pertain to a payment if the remaining time the payment might be locked for is larger than what the forwarding node would be willing to accept.
+This leakage could be evicted by adhering to the so-called shadow routes.
+
+One of the most subtle and yet powerful leakages an on-path adversary can foster is the timing analysis.
+An on-path adversary can log for every routed payment along with the amount of time it takes for a node to respond to an HTLC request.
+Before starting the attack, the attacker learns every node's latency characteristics in the Lightning Network by sending them requests.
+Naturally, this can aid in establishing the adversary's precise position in the payment path.
+Even more, as it was recently shown, an attacker can successfully determine the sender and the receiver of a payment from a set of possible senders and receivers using time-based estimators.
```
nopara73, line 198: Maybe make this statement timeless? You don't know when the reader will read this and unlike a research paper, this work evolves, so unless the reader has a git blame at hand he won't be able to tell what "recently" means.

```suggestion
Even more, an attacker can successfully determine the sender and the receiver of a payment from a set of possible senders and receivers using time-based estimators.
```

security_privacy_ln.asciidoc
```diff
@@ -0,0 +1,614 @@
...
+Therefore, Mallory can update her estimation of Alice's balance from "between zero and 1 million" to "between 500 thousand and one million."
+Another scenario happens if Alice's balance is lower than 500 thousand satoshis.
+In that case, Alice is unable to forward the payment and returns the "insufficient balance" error to Mallory.
+Mallory updates her estimation from "between zero and 1 million" to "between zero and 500 thousand."
+
+Note that in any case, Mallory's estimation becomes twice as precise after just one probing!
+She can continue probing, choosing the next probing amount such that it divides the current estimation interval in half.
+This well-known search technique is called _binary search_.
+With binary search, the number of probes is _logarithmic_ in the desired precision.
+For example, to obtain Alice's balance in a channel of 1 million satoshis up to a single satoshi, Mallory would only have to perform stem:[\log_2(1000000) \approx 20] probings.
```
nopara73, line 243: This formula doesn't get previewed here: https://github.com/seresistvanandras/lnbook/blob/lnsecurityprivacy/security_privacy_ln.asciidoc#linking-senders-and-receivers

nopara73 (commented):
Great chapter, I've learnt a lot from it! General suggestions:

- It would greatly help my understanding if it'd have a lot more illustration.
- It reads more like a research paper, rather than a book. I'm not sure if it's consistent with the rest of the book.

security_privacy_ln.asciidoc
seresistvanandras: Changed it to "preimage" everywhere.

security_privacy_ln.asciidoc
seresistvanandras: Indeed, you're right. Locally, in my text editor, it renders correctly. Maybe, it's a fault of GitHub? Anyone could confirm this? cc @s-tikhomirov  @carlaKC

security_privacy_ln.asciidoc
seresistvanandras: Added the appropriate references!

security_privacy_ln.asciidoc
seresistvanandras: Fixed

security_privacy_ln.asciidoc
seresistvanandras: Added the references

security_privacy_ln.asciidoc
seresistvanandras: Added a reference to https://1ml.com/statistics

security_privacy_ln.asciidoc
seresistvanandras: fixed

security_privacy_ln.asciidoc
seresistvanandras: fixed. added reference

security_privacy_ln.asciidoc
seresistvanandras: Thanks! Added your insight!

security_privacy_ln.asciidoc
seresistvanandras: fixed

seresistvanandras:
> Great chapter, I've learnt a lot from it! General suggestions:
> 
> * It would greatly help my understanding if it'd have a lot more illustration.

Absolutely agree. Would be nice if @Roasbeef @renepickhardt and @aantonop could tell us where it would be necessary to amend the text with more illustrations. Then, we could iterate on those figures right away before it gets too late.

ariard:
This is a great thread ! I wanna to do a review but I'm not sure about the breath/depth you're aiming for ?

The privacy section is pretty awesome. It's really thoughtful to talk about cross-layer linking (sadly you have far more heuristics than the ones already detailed...). Maybe a privacy framework should include LN nodes taxonomy (leafs, routing, merchants), it's likely you can observe a lot from their behavior patterns. Also risks of privacy leaks backfire on L1 coins, if utxos context-switches aren't well handled by your wallet.

On the security, it sounds limited to channel-jamming at the network level and doesn't mention other integrity concerns like channel/HTLC steals vectors, timevalue DoS, fees DoS, balance blackmailing, ... which affect link layer.


But ultimately, it's function of the public you're targeting ?
* node operators, who want to deploy their channels cautiously
* application devs, who want to inform their application design
* lightning devs, who have a hard time tracking the fast-evolving research
* base layer devs, who want a quick survey about L2 security & privacy
* curious users, who want to understand what LN is made of
* academic/security researchers, who want inspiration for their future work
* ...

And I guess that's only a small subset of the potential readers and their alleged motivations.

carlaKC:
> It reads more like a research paper, rather than a book. I'm not sure if it's consistent with the rest of the book.

Thanks for the feedback @nopara73! definitely something we want to address. Going to read through some of the other chapters to get an idea of tone/levelling.

> But ultimately, it's function of the public you're targeting ?

@ariad I'll defer to the main authors for authority, but from your list I would definitely rule out lightning+ base layer devs and researchers. Afaik, this book is not really intended for that level of specialization.

> doesn't mention other integrity concerns

In terms of other security concerns, it's difficult to outline them all in an approachable way, but I agree that this section could be fleshed out with flood&loot and a few other attacks. One concern I have is that some topics will be unfamiliar to a regular bitcoiner, for example transaction pinning, so we're going to have to go on a long tangent just to describe some of these attacks. 

This is already quite an content-dense chapter, so I think we're going to have to be satisfied with a disclaimer that the things we mention here are not exhaustive (privacy and security never can be, really).

security_privacy_ln.asciidoc
aantonop: This is a Github rendering limitation, but our asciidoc rendered will handle it correctly

aantonop:
> > But ultimately, it's function of the public you're targeting ?
> 
> @Ariad I'll defer to the main authors for authority, but from your list I would definitely rule out lightning+ base layer devs and researchers. Afaik, this book is not really intended for that level of specialization.
> 

The target audience for this book is a developer or computer science student. More specifically, technically-minded people with some background in Bitcoin who want to understand LN or develop applications that use LN. The first 4-5 chapters should also be accessible to non-developers but technical people with an interest in understanding the technology. Our target audience is not LN protocol developers or LN client developers.

aantonop:
What is the status of this PR? @s-tikhomirov ?

Have you addressed the comments from @Roasbeef ? Are we ready to merge this initial draft and subject it to further PRs for other editorial suggestions and changes?

seresistvanandras:
I only want to include a few cross-references across the chapter to increase readability. @Roasbeef requested a few clarifications on how certain inefficiencies and privacy leaks are mitigated by MPP payments. So, I think we are 98% ready for the merge. I'll add these latest this weekend hopefully.

Upon request, we would be happy to amend the accessibility of the chapter with graphs, figures, illustrations. However, we have not received yet any indications where such supplemental material would be useful, wanted. Let us know and then we can also add those figures and illustrations, if needed.

s-tikhomirov:
From my side, I have a deadline upcoming and can't allocate much time for this project for the next two weeks, unfortunately. I'll look at the state of this PR then and contribute if necessary.

nopara73:
A new paper is out on LN privacy: https://mobile.twitter.com/GeorgeKappos/status/1352969635679920131

https://arxiv.org/abs/2003.12470

seresistvanandras:
I think this PR is now ready to be merged. We addressed most of the comments and suggestions. Readability might be improved in the future by adding illustrations, figures and references to other chapters in the book.

PR #551, opened 2020-11-26
node_client typo

quantumcthulhu:


aantonop:
Honestly, I kinda like the idea of "make commandos". I imagine a Makefile in the hands of Rambo.

PR #555, opened 2020-11-29
P2P - Feature Bits

ImranLorgat:
Wrote a basic explanation for feature bits. Not 100% sure if I interpreted the string correctly. Will take another look and edit tomorrow.


ImranLorgat:
2nd commit: Expanded and rewrote large parts of this submission. I thought it would be more straightforward to break down a practical example and show readers exactly how a string of bits would be interpreted by another node.

I corrected an error from yesterday because I mistakenly did not realize that despite bit 2 not being used, bits 2 and 3 still form a "pair", so my count was off. One part I'm not 100% sure of is if I've read the string correctly. From right to left, I read bits 8 and 9 as "01", which is enabled and odd, so 'var_onion_optin' is enabled and backwards-compatible. Is this correct?

aantonop:
This is really superb work @ImranLorgat 

Thank you!

PR #563, opened 2020-12-09
Glossary - Feature Bits

ImranLorgat:


glossary.asciidoc
```diff
@@ -182,6 +182,12 @@ ephemeral key::
     This increases the security of transported messages or payments.
     Even if an ephemeral key leaks, only information about a single payment becomes public.
 
+feature bits::
+    A binary string that Lightning nodes use to communicate to each other which features they support.
+    Feature bits are included in many types of communication, such as invoices or channel announcements.
+    They can be decoded using BOLT #11, and will tell nodes which features the node has enabled, and whether these are backwards-compatible.
```
renepickhardt: shouldn't we refer to BOLT 9 as it includes a list of feature bits and also mention that those features are either local for a channel or global for the network?

renepickhardt (changes requested):
Concept ACK. I think it is important to include feature bits to the glossary but I believe BOLT 9 and the difference between global and local features are much more important

glossary.asciidoc
ImranLorgat: My apologies. Typo on my part.

BOLT 9 is the correct one and the one I referred to in the P2P chapter. BOLT 11 is for invoices! Fixed below.

PR #575, opened 2020-12-28
Consistency in spelling out the numbers that are < 10 (ch. 3)

jerzybrzoska:
It would be fine to maintain consistency in spelling out the numbers (provided that they are not code/math/ BTC amounts) that are < 10. I left out "chapter 6" and the like unchanged. 
[A post about style](https://www.grammarly.com/blog/when-to-spell-out-numbers/) for the reference.
>In scientific and technical writing, the prevailing style is to write out numbers under ten. While there are exceptions to these rules, your predominant concern should be expressing numbers consistently.

### DIFF
--- a/03_how_ln_works.asciidoc
+++ b/03_how_ln_works.asciidoc
@@ -22,7 +22,7 @@ A payment channel is a _financial relationship_ between two nodes on the Lightni
 
 The payment channel is managed by a _cryptographic protocol_, meaning a pre-defined process based on cryptography, used by the channel partners to re-distribute the balance of the channel in favor of one or the other channel partner. The cryptographic protocol ensures that one channel partner cannot cheat the other, so that the partners do not need to trust each other.
 
-The cryptographic protocol is established by the funding of a two-of-two _multi-signature address_ that requires the two channel partners to cooperate and prevents either channel partner from spending the funds unilaterally.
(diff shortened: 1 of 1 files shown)


03_how_ln_works.asciidoc
```diff
@@ -22,15 +22,15 @@ A payment channel is a _financial relationship_ between two nodes on the Lightni
 
 The payment channel is managed by a _cryptographic protocol_, meaning a pre-defined process based on cryptography, used by the channel partners to re-distribute the balance of the channel in favor of one or the other channel partner. The cryptographic protocol ensures that one channel partner cannot cheat the other, so that the partners do not need to trust each other.
 
-The cryptographic protocol is established by the funding of a 2-of-2 _multi-signature address_ that requires the two channel partners to cooperate and prevents either channel partner from spending the funds unilaterally.
```
bolatovumar: I believe 2-of-2 is a pretty standard spelling for these multigsignature wallet descriptions. Not sure if the rule of spelling out the digit should apply for these. I agree with other cases though (like `5` to `five`).

jerzybrzoska: You are right [Umar](@bolatovumar), I have not found a single source on the web with _two-of-two_, _2-of-2_ seems to be the standard. I already edited this PR, so now the changes do not include 2-of-2 modification. I find the issue of '2-of-2' vs 'two-of-two' interesting and may soon create a question on [English Stack Exchange](https://english.stackexchange.com/) in order to get their opinion.

bolatovumar (approved):

PR #580, opened 2021-01-04
ggch-channel-graph: initial draft of channel graph chapter

Roasbeef:
In this commit, we present an initial version of the chapter on the
channel graph itself, how to bootstrap a new peer to the network, how to
sync it, how to validate it, and how to maintain it once bootstrapped.

The current version of this chapter covers the following BOLTs: 07, 10.
BOLTs 09, 08, and 01 are alluded to along the way.

This is marked as a draft commit, as I still need to fill out the final
portion that covers how the current gossip network works, and the
protocol used to sync the channel graph based on block heights of
channels.

Due to the current length, we may want to consider eventually breaking
this chapter up, however I'm focusing on content first.

### DIFF
--- a/gossip.asciidoc
+++ b/gossip.asciidoc
@@ -1,0 +1,655 @@
+# Channel Graph Discovery, Authentication & Maintenance
+
+## Intro
+
+In this chapter, we'll explore exactly _how_ Lightning nodes discover each
+other, and also the channel graph itself. The channel graph is the
+interconnected set of publicly advertised channels (more on that later), and
+the nodes that these channels interlink. When most refer to the _network_ part
+of the Lightning Network, they're referring to the Channel Graph which itself
+is a unique authenticated data structured _anchored_ in the base Bitcoin
+blockchain.
+
+As the Lightning Network is a peer-to-peer network, some initial bootstrapping
+is required in order for peers to discover each other.  Within this chapter
+we'll follow the story of a new peer connecting to the network for the first
+time, and examine each step in the bootstrapping process from initial peer
+discovery, to channel graph syncing and validation.
+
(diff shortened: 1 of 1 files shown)


gossip.asciidoc
```diff
@@ -0,0 +1,655 @@
...
+;; Query time: 83 msec
+;; SERVER: 2600:1700:6971:6dd0::1#53(2600:1700:6971:6dd0::1)
+;; WHEN: Thu Dec 31 16:59:22 PST 2020
+;; MSG SIZE  rcvd: 138
+```
+
+In the above command, we've queried the server so we can obtain an `IPv4`
+address for our target node. Now that we have both the raw public key _and_ IP
+address, we can connect to the node using the `brontide` transport protocol at:
+`026c64f5a7f24c6f7f0e1d6ec877f23b2f672fb48967c2545f227d70636395eaf3@X.X.X.X`.
```
hfossli, line 215: It would be great if we can expand a little bit more here. How is this public key being used? I am assuming the public key is used for secure communcation, am I right? Can the "client" trust it has been given the correct public key going from the DNS? In a HTTP/browser world we never trust the dns, but always trust the CA's. I'm thinking there might be mitm-attacks here if the public key is used as a basis for KDF/encryption. What do you think @Roasbeef ?

hfossli: @aantonop and @Roasbeef, as I understand it this public key is used with ECDH and encryption handshakes https://github.com/lightningnetwork/lightning-rfc/blob/b201efe0546120c14bf154ce5f4e18da7243fe7a/08-transport.md#authenticated-key-agreement-handshake. Can the user fully trust that the unencrypted dns query is giving the correct public key and destination? How can we gain trust in this case?

Issue #528, opened 2020-10-17
Editorial:  Add muun wallet


klauslovgreen:
**Which part needs editing?**
Chapter 2 - wallets

**What editing does this need?**
Please consider adding muun wallet (https://muun.com)

**How does it currently read?**
Copy the text that needs editing, if it is small enough

**How should it read?**
Write a recommended solution here, if you can!

aantonop:
@klauslovgreen This is a great suggestion. Would you like to take a go at a pull request for this?

Fichte42:
The website states "Muun is a self-custodial wallet for bitcoin and lightning", but for the time being i added it as "Custodial" because of https://blog.muun.com/turbo-channels/

> It really depends on your preferences. When transactions require the use of turbo channels, there is a limited window of time where you trust Muun. Specifically, during that window of time, you trust Muun will not execute a - very public![1] - double-spend attack. If you are not okay with this, you should disable turbo channels.

aantonop:
Characterizing Muun as custodial because of the availability of turbo channels is incorrect and misleading, in my opinion. We should not change it. It is a non-custodial or self-custodial wallet.

Fichte42:
Ok, fixed it.

PR #586, opened 2021-01-18
Add Muun

Fichte42:
refs #528


aantonop:
Muun is not custodial.

PR #592, opened 2021-01-22
Changing buttlet point list to numbered list to match the below refer…

HannahMR:
Changing bullet point list to numbered list to match the below reference to question numbers.


renepickhardt (approved):
ACK

PR #596, opened 2021-01-23
Keeping fallback as one word for consistency.

HannahMR:
Keeping "fallback" as one word for consistency.


renepickhardt (commented):
need a native speaker to decide that. Intuitevly this seems wrong in that context but this is something that I cannot decide. @aantonop can you have a look please?

aantonop:
Rene is right about this. In this context we want to describe an action (verb) of falling back, not an attributive or adjectival noun of the result of the action (a fallback position).

PR #597, opened 2021-01-24
node_client.asciidoc: "collaboratively developed" instead of "collaborative developed"

jerzybrzoska:
[Collaborative](https://dictionary.cambridge.org/dictionary/english/collaborative) is not an adverb, [collaboratively](https://en.wiktionary.org/wiki/collaboratively) is. I also added 'are' for the sake of [parallelism](https://en.wikipedia.org/wiki/Parallelism_(grammar)), the said 'are' is optional, though.

### DIFF
--- a/node_client.asciidoc
+++ b/node_client.asciidoc
@@ -1,7 +1,7 @@
 [[set_up_a_lightning_node]]
 == Lightning Node Software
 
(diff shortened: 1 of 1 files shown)


renepickhardt (approved):
thanks!

PR #599, opened 2021-01-24
node_operations.asciidoc: the blockchain size update

jerzybrzoska:
Based on [the data](https://ycharts.com/indicators/bitcoin_blockchain_size). I myself run a pruned node so I cannot verify it. The adduced data may not include the transaction index. Another source that is consistent with the size that I propose: https://www.blockchain.com/charts/blocks-size


renepickhardt (commented):
I'd almost prefer to leave this one open until we deliver the book so that we can update at the very end

aantonop:
> I'd almost prefer to leave this one open until we deliver the book so that we can update at the very end

Please file an issue for long-running things we need to check before publishing. I don't want to leave open pull requests!

PR #605, opened 2021-01-26
Self-Custody-> Non-Custodial in table

jonki79:
The text above the table relates to non-custodial wallets, so to make it unambiguous, the wording was changed in the table.


aantonop:
I prefer the term Self Custody, as it defines a characteristic, not it's negation.

If we change something, we should change the preceding paragraph to maintain self-custody instead.

jonki79:
Fully agreed, regardless of what term is chosen, it should at least be consistent in text and table.

Issue #327, opened 2020-07-23
use eclair testnet?


keblek:
In ch2 the book throughly describes how to get some Bitcoin and load it into eclair. 

Currently [the fees](https://mempool.space) are like 120 sats/vByte and I don't want to compete for block space.

I understand that if you wait then they usually go back down, but I don't think that having to wait makes for a good learning experience if the reader wants to keep moving forward.

I am curious if there are any pedagogical reasons for diving straight into main net. 

Eclair has a testnet version of the app (which I assume is maintained in identical fashion to the mainnet version) which **should** be the same.

Thoughts? 

I guess it doesn't really matter (since its a drop in replacement) but if we are telling the user how to buy bitcoins maybe we should let them know that they can experiment with 0 risk since they are very new.

I think a small tip box that says you can use testnet with zero risk should suffice.

aantonop:
Great suggestion. @keblek do you want to take a stab at this with a pull request?

keblek:
Yes I can, see #612.

PR #612, opened 2021-01-31
Adding testnet wallet tip

keblek:
Added to let the reader know that this wallet is available and functions well enough to really test out lightning. Mainnet can be slow and expensive. #327


keblek:
#327

GustavoRSSilva (approved):

PR #613, opened 2021-01-31
changed git clone to HTTPS

keblek:
By using SSH on a fresh OS install you have to establish the public key of GitHub and if you don't have it you have to a song and dance about it. With HTTPS it works every time. Unless of course this is a security measure, in which case ignore this post.


aantonop:
Thanks @keblek !

Issue #615, opened 2021-01-31
Fix needed [c-lightning docker won't build]


keblek:
**Describe the error**
docker build doesn't work
![image](https://user-images.githubusercontent.com/68111648/106400019-3ec6a600-641c-11eb-93ae-dc404b374f42.png)

**To Fix**
Updating the build instructions in the docker script

**Additional context**

aantonop:
We test for specific versions of each LN client, because "latest" LND doesn't always work with "latest" Eclair and "latest" c-lightning. So for that reason, we pinned the versions.

Instead of unpinning, we need to move the versions forward for all tools, test that they continue to work across the entire example (setting up channels in a network and sending a payment).

keblek:
Yes I see your point, I tried to pin the latest clightning but I couldn't get the docker image to build until I deleted the version to make apt download the latest version.

What about adding an env file to centralise the pinning of the versions? We now have ubuntu, core, clightning, LND and eclair to track. I'm pretty sure this is possible.

Also, I think adding some guidelines to [CONTRIBUTING](https://github.com/lnbook/lnbook/blob/develop/CONTRIBUTING.md) to let testers know what constitutes a working version (minimum number of passing tests) would be helpful as well.

keblek:
#634 fixes this

keblek:
#640 really really fixes this

PR #616, opened 2021-01-31
Removing the version installs latest release

keblek:
#615


aantonop:
We want the versions pinned. If it is now broken, we need to figure out which version works for this example and pin it to that version.

PR #618, opened 2021-02-01
WIP: ch-channel-graph: fill out latter portion of chapter

Roasbeef:
In this commit, we pick up where we left off last and describe the
semantics of the `channel_update` message including how to validate it
and the significance of the message. We also describe in detail the usage
of the `channel_update` message within the routing network, which
includes an explanation of the fee structure, and the other fields
within the message itself.

To conclude the chapter, we describe the process used to _sync_ the
channel graph in the first place, as well as how peers maintain the
graph once it has been fully synced.

Similar to the prior commit, there's a lot of content in this commit that 
we may want to lift to other chapters once we finish the final ordering 
structure of the book itself.

### DIFF
--- a/gossip.asciidoc
+++ b/gossip.asciidoc
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
 ##### Channel Announcement Validation
 
 Now that we know what a `channel_announcement` contains. We can now move onto
-to exactly _how_ to verify such an announcement.
+to exactly _how_ to verify such an announcement. Verifying a channel
+announcement entails verifying that:
 
+ 1. The `chain_hash` specified by the channel announcement belongs to a known,
+ and supported blockchain.
+
+ 2. An _unspent_ UTXO exists in the chain as identified by the short channel
+ ID.
+
+ 3. The unspent UTXO has a p2wsh multi-sig script that is composed of
+ `bitcoin_key_1`, and `bitcoin_key_2`, ordered in lexicographic order.
+
+ 4. The signatures of `node_signature_1`, `node_signature_2`,
+ `bitcoin_signature_1`, and `bitcoin_signature_2` are valid signatures under
+ the corresponding public keys over the defined message digest.  The message
+ digest is the serialization of the raw message bytes after the final
+ signature. More on this in the chapter that covers wire message framing.
+
+The channel announcement is a core component of the Lighting Network channel
+graph, as it itself is an unauthenticated proofs of the existence of a channel.
+As verification of channel announcement ensures that the output is _unspent_ in
+order to add a new edge (a channel) to the public channel graph an upfront
+chain fee must be paid. In addition to this by verifying the signatures of both
+the bitcoin and node signatures, we ensure that the parties advertising the
+channel are able to update the underlying multi-sig of the channel, and also
+are aware of each other (all pubkeys are signed).
+
+A channel announcement is made available to the public network only for the set
+of _public_ channels that wish to be used for forwarding/routing payments. Only
+a single instance of a channel announcement exists for a given channel. The
+channel announcement describes the "base" of a channel on the network (it's
+total capacity, location, participants, etc), but not the _shape_ of the
+channel. A prospective sender needs to know bout the shape and the shape in
+order to route through a channel. 
+
+In our terminology, the "base" of a channel is fixed and cannot be changed
+(only by spending the output which closes the channel). On the other hand, the
+"shape" of the channel can be changed simply by updating the information that
+advertise _how_ to route through the channel. When we say "how" , we refer to
+what fees to use, the largest payment accepted, etc. As the channels are
+bi-directional, that routing policy (how one should route through the channel)
+needs to be defined in both directions. We call this routing policy the
+`channel_update`, which both sides each being able to indecently update the
+routing policy for "their end" of the channel.
 
 #### Channel Update: Structure & Validation
 
+Now that we've covered the message that advertises the base (the unchangeable
+attributes that anchor a channel to the main chain) of a channel, we'll cover
+the advertisement that governs the "shape" of the channel, or its routing
+policy. The routing policy for a given channel announcement is referred to as
+the `channel_update` for the channel.
 
-// TODO(roasbeef): rename to "the structure of the channel graph"?
+Similar to the `channel_announcement`, then `channel_update` is authenticated
+via a digital signature, to ensure that only the party that actually _owns_ the
+channel is able to update the routing policy for said channel. This prevents
+other parties from spoofing "fake" channel updates. Unlike
+`channel_announcement`s which are created, then set it stone, a
+`channel_update` can be updated at any time by either end of a channel. As
+we'll see below, nodes in the network employ spam prevention and throttling
+heretics to ensure the update traffic doesn't overwhelm the network
+
+With the basics, covered, we'll now explore the precise structure of the
+`channel_update` and how each of the encoded fields are used by senders to
+route through a channel, and also how a forward ng node can "classify" its
+channel to tend towards certain types of traffic based on its selection of the
+`channel_update` parameters.
+
+##### Anatomy of a Channel Update
+
+The channel update is the lifeblood of the core of the routing network.
+Advertising a new `channel_update` allows routing nodes to modify their fees,
+the status of their channel, security parameters, and other critical routing
+information.
+
+First, we'll enumerate each field, and provide a brief description. In a later
+section, we'll examine the fields that dictate a channel's directional routing
+policy in more detail.
+
+  * `signature`: A valid ECDSA signature over the `channel_update` message
+    digest signed by the node that's advertising this update.
+
+  * `chain_hash`: A 32-byte hash that denotes the chain that this channel is
+    anchored in. This is typically the genesis hash of the chain.
+
+  * `short_channel_id`: The 64-bit short channel ID that identifies which
+    channel in the Channel Graph is being updated.
+
+  * `timestamp`: A 32-bit time stamp typically interpreted as a unix epoch
+   timestamp which allows an update to create a relative sequencing of their
+   channel updates. An update with a higher timestamp is more recent than one
+   with a lower one.
+
+  * `message_flags`: A bitfield that's used to denotes the presence of optional
+   fields.
+
+  * `cltv_expiry_delta`: A forwarding related security parameter for the
+   channel. This value denotes the minimum gap (or delta) between the incoming
+   and outgoing HTLCs that are to be routed thorough this channel.
+
+  * `htlc_minimum_msat`: A value that denotes the smallest HTLC that's
+   permitted to be routed through this channel. This is expressed in
+   milli-satoshis.
+
+  * `fee_base_msat`: The _base_ fee that must be paid for all payments routed
+   through this node regardless of payment size. This is expressed in
+   milli-satoshis.
+
+  * `fee_proportional_millionths`: The _proportional_fee to be paid in
+   millionths of milli-satoshi. This is the denominator to be used in _fixed
+   point_ arithmetic when computing the _rate_ that needs to be paid when
+   routing through a channel.
+
+  * `htlc_maximum_msat`: The largest HTLC that is permitted to be routed
+   through this channel. This is expressed in milli-satoshis.
+
+Now that we know all the fields of the `channel_update` message, we'll describe
+at a high level how verification of a channel update is to be carried out. Most
+importantly, the `signature` MUST be valid, and the `timestamp` MUST always be
+increasing in order to invalidate prior updates.
+
+##### Channel & Message Flags
+
+As mentioned earlier, there're two `channel_update` messages for each direction
+of the channel. Therefore, we need to be able to identify which direction of
+the channel is being updated. If we have two nodes Alice, and Bob, then there
+may exist a channel update in the direction of Alice -> Bob, as well as for Bob
+-> Alice. In the first case, we require Alice to be the one that signs the
+update (as she will be carrying the HTLC), and in the second case, we require
+that Bob is the one that signs the update.
+
+Thankfully, we can use a single bit to denote which direction is sending out
+the update. Recall that in the `channel_announcement` message, we sort the node
+public keys of both nodes lexicographically and refer to the "first" node as
+`node_1`, and the second as `node_2`. We then carry this over to the
+`channel_update` message within the `channel_flags` field. The field itself is
+a bit field, with the LSB denoting the "direction" of the channel update. If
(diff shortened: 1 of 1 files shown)


gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+order to add a new edge (a channel) to the public channel graph an upfront
+chain fee must be paid. In addition to this by verifying the signatures of both
+the bitcoin and node signatures, we ensure that the parties advertising the
+channel are able to update the underlying multi-sig of the channel, and also
+are aware of each other (all pubkeys are signed).
+
+A channel announcement is made available to the public network only for the set
+of _public_ channels that wish to be used for forwarding/routing payments. Only
+a single instance of a channel announcement exists for a given channel. The
+channel announcement describes the "base" of a channel on the network (it's
```
renepickhardt, line 677: I am not sure about the terminology of `base` and `shape` of a channel. 

In any case I suggest to make the definition more explicit if we need it instead of introducing it ad hoc. 

Also a summary / list of features that a channel has might be useful: 

* base: capacity, location, participants
* shape (I used to call that meta data): fees, forwarding policies, cltv,...
* state:  balance, htlcs in limbo, active / online,  {opening, open, closing, closed}

renepickhardt: maybe we could have: 

* constant channel parameters (formerly base)
* dynamic channel parameters (formaly shape / meta data)
* channel state (which could be divided into public (e.g. if the channel is open) and unpublished state (e.g.: balance information))

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+ ID.
+
+ 3. The unspent UTXO has a p2wsh multi-sig script that is composed of
+ `bitcoin_key_1`, and `bitcoin_key_2`, ordered in lexicographic order.
+
+ 4. The signatures of `node_signature_1`, `node_signature_2`,
+ `bitcoin_signature_1`, and `bitcoin_signature_2` are valid signatures under
+ the corresponding public keys over the defined message digest.  The message
+ digest is the serialization of the raw message bytes after the final
+ signature. More on this in the chapter that covers wire message framing.
```
renepickhardt, line 663: I really like including the 4 validation steps into the text. however I wonder if it makes sense if we would actually include the hex of an announcement message, dissect it and run through all steps as an example. It is a lot of work and it seems a bit random to do it for the channel announcement but I guess it is reasonable to do it somewhere and the longer I think about it the more instructive I find it. For example I would not find it that helpful with messages from BOLT 02

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+The channel announcement is a core component of the Lighting Network channel
+graph, as it itself is an unauthenticated proofs of the existence of a channel.
+As verification of channel announcement ensures that the output is _unspent_ in
+order to add a new edge (a channel) to the public channel graph an upfront
+chain fee must be paid. In addition to this by verifying the signatures of both
+the bitcoin and node signatures, we ensure that the parties advertising the
+channel are able to update the underlying multi-sig of the channel, and also
+are aware of each other (all pubkeys are signed).
+
+A channel announcement is made available to the public network only for the set
```
renepickhardt, line 674: Maybe splitting into 

* creation of channel announcement message (and put that towards the end of the channel construction chapter) 
* validation as mostely done here and keep it in the gossip chapter

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+channel is able to update the routing policy for said channel. This prevents
+other parties from spoofing "fake" channel updates. Unlike
+`channel_announcement`s which are created, then set it stone, a
+`channel_update` can be updated at any time by either end of a channel. As
+we'll see below, nodes in the network employ spam prevention and throttling
+heretics to ensure the update traffic doesn't overwhelm the network
+
+With the basics, covered, we'll now explore the precise structure of the
+`channel_update` and how each of the encoded fields are used by senders to
+route through a channel, and also how a forward ng node can "classify" its
```
renepickhardt, line 711: nit: forwarding

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+increasing in order to invalidate prior updates.
+
+##### Channel & Message Flags
+
+As mentioned earlier, there're two `channel_update` messages for each direction
+of the channel. Therefore, we need to be able to identify which direction of
+the channel is being updated. If we have two nodes Alice, and Bob, then there
+may exist a channel update in the direction of Alice -> Bob, as well as for Bob
+-> Alice. In the first case, we require Alice to be the one that signs the
+update (as she will be carrying the HTLC), and in the second case, we require
```
renepickhardt, line 775: we have two different definitions of channel direction: 

1. DER encoding of lesser node public key
2. actual flow: Who offers an HTLCs. 

When I first learned about Lightning Networkwork I was very confused why the routing policy on a channel is defined by the node who adds the outgoing HTLCs and not why the routing fee is already for an incoming HTLC on my channel  

We mentioned implicitely here that there is this the direction with the outgoing HTLC. I think this should be defined somewhere very explicitely or appended with some info box doubeling down about the direction definitions

renepickhardt: I assume the "direction" of channels will already be mentioned in routing

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+that Bob is the one that signs the update.
+
+Thankfully, we can use a single bit to denote which direction is sending out
+the update. Recall that in the `channel_announcement` message, we sort the node
+public keys of both nodes lexicographically and refer to the "first" node as
+`node_1`, and the second as `node_2`. We then carry this over to the
+`channel_update` message within the `channel_flags` field. The field itself is
+a bit field, with the LSB denoting the "direction" of the channel update. If
+the bit is set, then this update belongs to `node_1`, otherwise to `node_2`.
```
renepickhardt, line 784: Ofc we describe it how it is but I am wondering: Couldn't this be also resolved by looking who signed the channel update message?

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+outgoing and incoming HTLCs of a forwarded payment. By sending out less than
+they receive, routing nodes are able to profit by earning a fees for their help
+to route the HTLC through the network. All HTLCs in the LN are expressed in
+`msat` or milli-satoshis instead of satoshis or Bitcoin. We use `msat`, as it
+allows us to charge fees for 1 sat payments. The routing fees has two
+component, `fee_base_msat` and `fee_proportional_millionths`. In order to
+calculate how much a payment of `N msat` should pay to route through a channel,
+we use the following equation:
+```
+fee = fee_base_msat + (N * fee_proportional_millionths) / 1,000,000
```
renepickhardt, line 804: That is somewhat redundant with the routing chapter where we validate against it.

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
 
-// TODO(roasbeef): rename to "the structure of the channel graph"?
+This may look a bit strange at first, but all we're doing in the latter half of
+the equation is computing the rate to be paid. Rather than floating point
+arithmetic, we use fixed point arithmetic with 1 million as our base. This lets
+us express very small amounts of fees to be paid in the `mast` unit. Let's say
+we use a value of `100,000` for `fee_proportional_millionths`, this is
+equivalent to a `10%` fee as `100,000/1,000,000 = 0.10` or `10%.
+
+Another important security related parameter for a channel is the
```
renepickhardt, line 814: I would move this somewhere. Either we move it to the security chapter or routing or make it an info box. 

In any case I think we should also create an example where we show with specific timelocks and blockheights / mempool what could happen. 

I think something like this is one of the more crucial things for a student to understand thus my suggestion to heavily elaborate on this

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+payments. If the second LSB bit is `1`, then the channel is disabled, otherwise
+it's enabled.
+
+Next, we have the `htlc_minimum_msat` and `htlc_maximum_msat` fields. As their
+name entails, this allows a node operator to restrict the _size_ of HTLCs that
+are permitted to be routed through the channel. Setting these values can
+recreate a sort of "routing profile" for the channel. As an example, if we set
+our `htlc_minimum_msat` to a very high value, then we signal that this channel
+isn't to be used to route micropayments. If we set our `htlc_maximum_msat` to a
+very low value, then we signal the converse, instead that this channel is only
```
renepickhardt, line 842: should we discuss the issue with dustoutputs? 

we also did not discuss the `max accepted htlc` parameter. Even if it is not part of gossip it should be discussed as a security thing as other things that happen in gossip. 

Bot topics could be info boxes

renepickhardt (changes requested):
Finnished up until ` Syncing the Channel Graph`

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+  * `chain_hash` : The genesis hash of the chain we're querying for channels
+    within.
+
+  * `first_block_num`: The first block in the range we're interested in short
+    channel IDs of.
+
+  * `number_of_blocks`: The number of blocks after `first_block_num` that the
+    query should encompass. This is the end point of our query.
+
+  * `query_channel_range_tlvs: A series of optional TLV fields that are used to
```
renepickhardt, line 893: I guess for a reader it is confusing how to understand why the `query_channel_range` message does not have a length field in comparison to other messages. How does the node know if the `query_channel_range_tlvs` is empty or has elements? I assume we describe this in the messaging and framing chapter as TLV format belongs there I am just mentioning this here to potentially set a cross reference?

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+start at a low blockheight say 1000, we can then use a lower bound (0 in this
+case) to "seek" through the channel graph. We'll update our min and max block
+height cursors in order to request all the `short_channel_id`s that exist in
+that range. Once we receive that, we can then ask a  node for the precise
+`channel_announcement`, `channel_update`, and `node_announcement` that allows
+us to insert that edge (and the vertexes) into our local channel graph.
 
+The protocol we just described above at a high level is used to allows nodes to
+sync all the known (active) channels within the Channel Graph from a particular
+block range onwards. Within the protocol itself, we use two pairs of messages
```
renepickhardt, line 863: why is that useful? I imagine as a new node I would just say give me between 0 and current blockheight. 

What I try to say until today I do not know why the protocol was designed in this way and it is not mentioned here either.  can we elaborate 1 or two sentences about this choice?

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+
+  * `chain_hash`: The genesis hash of the chain that short channel IDs are
+    being returned for.
+
+  * `first_blocknum`: The first block number in the range of replies.
+
+  * `number_of_blocks`: The number of blocks after the frist block number that
+    channel IDs are being returned for.
+
+  * `full_information`: A byte that denotes if the sending node has all
```
renepickhardt, line 912: how does a node know this? I think the protocol is only eventually consistant. So if you connect to me I send you the answer I have full information but leave out channels you would also believe to have full information. 
I understand that light nodes are aware of the fact that they might not have full information but is the other way around even decidable?

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+    we're querying for channels within.
+
+  * `len`: The number of bytes in the remaining payload.
+
+  * `encoded_short_channel_ids`: The set of encoded short channel IDs.
+
+  * `query_short_channel_ids_tlvs`: A set of optional TLV fields that are used
+    to extend the query system.
+
+Similar to the prior message in structure, this allows the syncing node toe
```
renepickhardt, line 956: NIT: to

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+request that the responding node send over the full information of the
+referenced short channel IDs, as they're missing from its local database.
+
+Upon receiving this message, the responding node should send over all
+`channel_announcement`, `channel_update`, and `node_announcement` messages (in
+that order) that are related to the reference short channel IDs. Note that the
```
renepickhardt, line 962: why do the node announcements come last? that seems counter intuitive

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+
+  * `query_short_channel_ids_tlvs`: A set of optional TLV fields that are used
+    to extend the query system.
+
+Similar to the prior message in structure, this allows the syncing node toe
+request that the responding node send over the full information of the
+referenced short channel IDs, as they're missing from its local database.
+
+Upon receiving this message, the responding node should send over all
+`channel_announcement`, `channel_update`, and `node_announcement` messages (in
```
renepickhardt, line 961: I assume the replying node will only send the most recent channel_update message? If so we should state this explicitely

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+component of the system as it allows routing nodes to update their routing
+policies due to changes in the market, chain conditions, or simply their own
+preference.
 
-### Purpose
-### Structure
-### Validation
+Once we're fully send the graph, we can now being to receive any new node
+announcements, channel announcements, or channel updates via the _gossip
+network_. By gossip network we simply mean a set of peers that rumor updates to
```
renepickhardt, line 997: it seems to me that this part should be more at the beginning of the chapter. it actually explains the gossip nature of the protocol. maybe we could flip the ongoing channel graph maintenance and the initial sync

gossip.asciidoc
```diff
@@ -644,68 +644,388 @@ channel graph) has the following attributes:
...
+the Channel Graph, in order to _eventually_ propagate their updates to the
+entire network. 
 
-### Channel Updates
+It's important to note that peers on the network _won't_ immediately being to
```
renepickhardt, line 1001: why not spam protection or as a result of the timstamp_filter mechanism that can be used as discussed in the last paragraph

renepickhardt (changes requested):
Over all very nice! I think one can understand the protocol from the text. 
Though I think it occasionally lacks motivation and why which could be either added or in case of space limitations exchanged with some of the very technical details.

aantonop:
@Roasbeef status of this PR? Are we ready to merge?

aantonop:
I picked up the most relevant parts of this PR in the rewrite of gossip chapter.

Issue #590, opened 2021-01-22
Editorial: Consider replacing "Wei" nomenclature


AndrejMitrovic:
**Which part needs editing?**

https://github.com/lnbook/lnbook/blob/develop/routing.asciidoc

**What editing does this need?**

Replace "Wei" with "Charlie" or with any other name that happens to start with a letter somewhere between Bob and David / Gloria.

The reasoning I have for this is it makes the names sorted alphabetically, which makes it easier to track the "hop index" in the payment channel. Right now it reads "alice, bob, wei, david", and it's hard to memorize that "wei" is the 3rd hop. It would be much easier if it was Charlie or Chris / Christina, or something similar. E.g. Alice / Bob / Charlie / David. 

-----

Also the graphs are showing Gloria as the destination node even though the text refers to it as David. Should I file another issue?

keblek:
#517

PR #623, opened 2021-02-02
WIP: Rename story actors: Fixes #464 #517 #590

### LINKED ISSUE #464: Editorial: Use alphabetical placeholder names in examples
**Which part needs editing?**
The routing chapter goes somewhat against the [convention](https://en.wikipedia.org/wiki/Alice_and_Bob) by using the placeholder names that do not start with the first letters of the alphabet: Alice - Bob - Wei - Gloria. The reader may have to switch back and forth between the schemes and the text to remember whether e.g. Wei comes before or after Gloria.

**What editing does this need?**
Use standard alphabetical names (e.g. Alice - Bob - Charlie - David).

### LINKED ISSUE #517: Editorial: Wei name potentially confusing
The name `Wei` may be confused with the [Ethereum Wei definition](https://github.com/ethereumbook/ethereumbook/blame/develop/glossary.asciidoc#L312). The reader may obviously understand the context or be unfamiliar with that definition, but still for some readers it may create confusion in json examples or slow readability.

**Which part needs editing?**
Through all the book.

**What editing does this need?**
Simplify, Re-name.

aantonop:
CAUTION EXTENSIVE CHANGES/POTENTIAL CONFLICTS: This pull request will introduce changes across the entire book. 

The examples we built based on user stories have caused some confusion. 

The existing examples use Alice, Bob, Wei and Gloria, first introduced in chapter 1.

When picking which participants make sense in each story, we broke the alphabetic ordering of names A->B->C->D. We have Alice and Bob to set the stage, but the following names are no longer alphabetically ordered, causing confusion as to the path of the payment. Problem identified in #464 

Additionally, the name "Wei" may cause confusion for Ethereum users, as it has been used to name the smallest ETH unit. Problem identified in #517 #590

When changing the names to maintain alphabetic ordering, we are rejecting the suggested "Alice, Bob, Charlie, David" names, because they lack geographic/language/gender diversity that we have intentionally introduced in all our stories. 

We are changing the name Wei to **Chan**
We are changing the name Gloria to **Dina**

The examples will now reference a network of payment channels flowing like this:

Alice -> Bob -> Chan -> Dina, with Alice making a payment to Dina.


GustavoRSSilva (approved):

AndrejMitrovic:
Thanks a lot for your work!

What about the graphics though? They still seem to use the old names.

aantonop:
Yes, it's a work in progress. All the illustrations will be re-done.

PR #626, opened 2021-02-07
Added -r option to jq to strip quotes

keblek:
The manual way for stripping quotes from a json reply is messy and gets in the way, jq has a built in tool `-r` to do this.

Aligned the echo commands so its a bit more tidy

### DIFF
--- a/code/docker/setup-channels.sh
+++ b/code/docker/setup-channels.sh
@@ -1,29 +1,15 @@
 #!/bin/bash
 
 echo Getting node IDs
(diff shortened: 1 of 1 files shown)


aantonop:
Nice! I (clearly) didn't know that and I appreciate learning about it. I was looking for such an option in jq and couldn't find it. It was very frustrating to have to resort to shell substitution!

Thank you

keblek:
You're very welcome! I'm also learning a lot from reading this book.

PR #633, opened 2021-02-08
Bump Bitcoin Core to latest version

keblek:
This latest version has signet so I think it will benefit lightning users since signet is a better testnet for lightning than testnet3.

New default wallet isn't created at start


michaelfolkson:
ACK on Signet. Signet has a simulated Poisson process for block discovery imitating mainnet which should be useful for timelock experimentation (in comparison to testnet's highly erratic block discovery)

PR #634, opened 2021-02-08
Bump clightning to 9.3

keblek:
also removed extra cli argument `--network regtest` clightning doesn't use it, but it will happily ignore it


keblek:
I added more things in addition to the bump to c lightning.

There is now an extra docker example called `doughtnuteconomy` which sets up six clightning nodes all connected to each other that pay each other in a loop. 

Just as an example for how to spin up a bunch of nodes at the same time.

tompro:
This seems to be important as building containers locally with c-lightning 9.2 currently fails. For some reason it is no longer in the PPA.

code/perpetualpayments/Makefile
```diff
@@ -0,0 +1,16 @@
+
+REGISTRY=docker.com
+NAME=lnbook
+CONTAINERS=bitcoind c-lightning
+
+all: build-all push-all
```
tompro: Had a try today, would'nt it be better to default make to just create the containers and require an argument if it should be pushed to the registry. Don't know whether this will be part of the book resources though.

keblek: I’m gonna delete the extra stuff I added and make another PR. The bump is necessary.

aantonop:
Let's please separate these two:

- Bumping c-lightning to get the existing examples to work
- Adding a completely new script and example

Please submit these as two separate PRs. As it is, I can't merge.

PR #640, opened 2021-02-18
Bump c-lightning to 9.3

keblek:
Originally this was just a c lightning change but then I added an example docker script.

I have [removed that example script](https://github.com/lnbook/lnbook/commit/32b3480e4204f2a81cd35f778f7c2b8c4d71ff8f) and this is just the clightning bump to replace #634.


code/docker/c-lightning/bashrc
```diff
@@ -1,4 +1,4 @@
-alias lightning-cli="lightning-cli --network regtest --lightning-dir=/lightningd"
+alias lightning-cli="lightning-cli --lightning-dir=/lightningd"
```
renepickhardt, line 1: why would you remove `regtest` network?

code/docker/c-lightning/c-lightning-entrypoint.sh
```diff
@@ -6,7 +6,7 @@ source /usr/local/bin/wait-for-bitcoind.sh
 echo Starting c-lightning...
 lightningd --lightning-dir=/lightningd --daemon
 
-until lightning-cli --lightning-dir=/lightningd --network regtest getinfo > /dev/null 2>&1
+until lightning-cli --lightning-dir=/lightningd getinfo > /dev/null 2>&1
```
renepickhardt, line 9: same here. why removing `regtest`?

renepickhardt (dismissed):
I thought we purposly do everything in regtestmode. so I am a little bit confused by thouse changes

code/docker/c-lightning/bashrc
aantonop: I think it defaults to regtest because of the config. I will test

code/docker/c-lightning/c-lightning-entrypoint.sh
aantonop: I think it defaults to regtest because of the config, I will test

code/docker/c-lightning/bashrc
keblek: @rene78 lightning-cli does not accept regtest as a parameter for network, it does accept basically all the other networks. However, it does not complain about it either which I think is weird. 

https://github.com/ElementsProject/lightning/pull/4374#discussion_r571752449

PR #649, opened 2021-03-13
Alphabetize list terms, add hash function, & heading

pamelawjd:
Alphabetized terms under "Lightning Network Basic Concepts" and added Hash Function and Digital Signature as terms because they are also used in this chapter. Used glossary definitions (partial) to define in this section. 
Added 2nd level heading to Scaling Blockchains and added italics to what if statement at the end of the section, which is the transition to LN.


renepickhardt (approved):
LGTM

PR #666, opened 2021-03-30
Format change preface.asciidoc

ghost:
Addresses don't use bullet points. 

(I am new to this, so I'm not sure it two spaces at the end of a line is the best way to indicate a line break.)  I'm planning on reading the whole book - I have quite a lot of experience of editing for publication.  Expect more from me.


renepickhardt (commented):
hey @aantonop can you say something about this. I assume there was a reason why it was in this way?

aantonop:
This is O'Reilly content and it is included as they have formatted it for the Atlas platform.

PR #667, opened 2021-03-30
Update glossary.asciidoc

ghost:
Minor formatting changes.
I've also reworded some definitions to make them clearer and more consistent.
In three cases, I removed additional sentences from entries because they added confusion rather than clarity.
There is still a bit more checking of the Glossary to be done.


glossary.asciidoc
```diff
@@ -92,63 +92,61 @@ Breach Remedy Transaction::
...
-    A special field used as the sole input for coinbase transactions. The coinbase allows claiming the block reward and provides up to 100 bytes for arbitrary data.
+    A special field used as the sole input for coinbase transactions. The coinbase allows the block reward to be claimed and provides up to 100 bytes for arbitrary data.
     The block reward consists of two things.
     First, newly generated coins. The amount of allowed coins to be generated is part of the consensus rules and decreases over time based on the current block height.
     Second, the miner is also allowed to add all the fees of the transactions from the current block to the coinbase.
     Not to be confused with coinbase transaction.
 
 coinbase transaction::
-    The first transaction in a block. Always created by a miner, it includes a single coinbase.
+    The first transaction in a block which is always created by a miner and which includes a single coinbase.
```
giovannizotta, line 118: I'm not sure if the modified version is better than the original here.

glossary.asciidoc
```diff
@@ -92,63 +92,61 @@ Breach Remedy Transaction::
...
-Diffie Hellman Key Exchange::
+Diffie Hellman (DH) Key Exchange::
     On the Lightning Network, the Elliptic Curve Diffie-Hellman method is used.
     It is an anonymous key agreement protocol that allows two parties, each having an elliptic-curve public-private key pair, to establish a shared secret over an insecure communication channel.
     This shared secret may be directly used as a key, or to derive another key.
     The key, or the derived key, can then be used to encrypt subsequent communications using a symmetric-key cipher.
-    An example of the derived key would be the shared secret between the ephemeral session key of a sender of an onion with the nodes public key of a hop of the onion as described and used by the SPHINX Mix Format.
-    Via https://en.wikipedia.org/w/index.php?title=Elliptic-curve_Diffie%E2%80%93Hellman&oldid=836070673
-
+
```
giovannizotta: I understand the reasoning behind the thought of removing this, however I'm not sure if it's right thing to do. If the person reading doesn't want to dive deep, they would just skip the link. Since it's in the glossary, I don't think it clogs up the book

glossary.asciidoc
```diff
@@ -308,7 +304,7 @@ Noise_XK::
...
+    More particularly (from: http://www.noiseprotocol.org/noise.html), the protocol enables encryption to a known recipient and strong forward secrecy. This payload is encrypted based on an ephemeral-ephemeral DH as well as an ephemeral-static DH with the recipient's static key pair. Assuming the ephemeral private keys are secure, and the recipient is not being actively impersonated by an attacker that has stolen its static private key, this payload cannot be decrypted. Sender authentication is resistant to key-compromise impersonation (KCI). The sender authentication is based on an ephemeral-static DH ("es" or "se") between the sender's static key pair and the recipient's ephemeral key pair. Assuming the corresponding private keys are secure, this authentication cannot be forged.
```
giovannizotta: Maybe since we're changing this, it would be worth to refactor the first sentence (maybe it's just me, but I don't really like "more particularly") with something like this: "In particular, the protocol (from: http://www.noiseprotocol.org/noise.html) enables encryption and strong forward secrecy to a known recipient."

glossary.asciidoc
```diff
@@ -341,18 +336,18 @@ P2WPKH::
     The signature of a P2WPKH (Pay-to-Witness-Public-Key-Hash) contains the same information as a P2PKH spending, but is located in the witness field instead of the scriptSig field. The scriptPubKey is also modified.
 
 P2WSH::
-    The difference between P2SH and P2WSH (Pay-to-Witness-Script-Hash) is about the cryptographic proof location change from the scriptSig field to the witness field and the scriptPubKey that is also modified.
+    The difference between P2SH and P2WSH (Pay-to-Witness-Script-Hash) is the location change of the cryptographic proof from the scriptSig field to the witness field and the modification of the scriptPubKey.
 
 payment::
-    A payment occurs if we transfer bitcoin within the Lightning Network.
+    A payment occurs if bitcoin is stransferred within the Lightning Network.
```
giovannizotta: There is a typo here

glossary.asciidoc
```diff
@@ -423,13 +418,13 @@ Satoshi Nakamoto::
...
 Script::
-    Bitcoin uses a scripting system for transactions. Forth-like, Script is simple, stack-based, and processed from left to right. It is purposefully not Turing-complete, with no loops.
+    Bitcoin uses a scripting system for transactions called Script. Forth-like, it is simple, stack-based, and processed from left to right. It is purposefully not Turing-complete, with no loops.
 
 ScriptPubKey (aka pubkey script)::
     ScriptPubKey or pubkey script, is a script included in outputs which sets the conditions that must be fulfilled for those satoshis to be spent. Data for fulfilling the conditions can be provided in a signature script.
 
 ScriptSig (aka signature script)::
-    ScriptSig or signature script, is the data generated by a spender which is almost always used as variables to satisfy a pubkey script.
+    ScriptSig or signature script is the data generated by a spender, which are almost always used as variables to satisfy a pubkey script.
```
giovannizotta, line 430: I think the subject of the sentence is a bit confusing. Shouldn't this all be singular? Like: "ScriptSig or signature script is the data generated by a spender, which is almost always used as a condition to satisfy a pubkey script."

glossary.asciidoc
```diff
@@ -470,15 +465,15 @@ soft fork::
 
 SPHINX Mix Format::
     A particular technique for Onion Routing used in the Lightning Network and invented by George Danezis and Ian Goldberg in 2009.
-    With the SPHINX Mix Format, each message of the onion package is padded with some random data so that no single hop can estimate how far on the route they are.
+    With the SPHINX Mix Format, each message of the onion package is padded with some random data so that no single hop can estimate how far along the route it has traveled.
```
giovannizotta, line 471: ... "how far along the route each message has traveled" maybe? I'm not sure

glossary.asciidoc
```diff
@@ -470,15 +465,15 @@ soft fork::
...
+Submarine Swap::
+    A Submarine Swaps enables transfers between on-chain Bitcoin addresses and off-chain locations, like the Lightning Network. Just as standard LN transfers chain payments by means of HTLCs that make the final claim on funds conditional on the recipient revealing a secret to all links in the chain, Submarine Swaps use the same logic and procedure to transfer funds across the on-chain/off-chain barrier with minimal trust. They can also be used to enable transfers from another chain, say Litecoin, to an off-chain LN address. Reverse Submarine Swaps allow bitcoin transfers in the opposite direction, from an off-chain LN location to an on-chain address.
```
giovannizotta, line 476: Small typo: at the beginning of the sentence there is "A Submarine Swap**_s_** enables ... ", while it should be "A Submarine Swap enables ..."

glossary.asciidoc
```diff
@@ -470,15 +465,15 @@ soft fork::
...
 
-timelocks::
-    A timelock is a type of encumbrance that restricts the spending of some bitcoin until a specified future time or block height. Timelocks feature prominently in many Bitcoin contracts, including payment channels and hashed timelock contracts.
+timelock::
+    A timelock is a type of encumbrance that restricts the spending of some bitcoin until a specified future time or block height. Timelocks feature prominently in many Bitcoin contracts, including payment channels and Hashed Timelock Contracts.
```
giovannizotta, line 479: I believe it should be "Hash Time Locked Contracts"

glossary.asciidoc
```diff
@@ -488,13 +483,13 @@ transaction::
...
+    Transaction malleability is a property that the hash of a transaction can change without changing the semantic of the transaction (the UTXOs it is spending, the destinations and the corresponding amounts).
     For example, altering the signature can change the hash of a transaction, because of the non-deterministism of ECDSA signing.
-    A commitment transaction needs the hash of a funding transaction, if the hash of the funding transaction changes, transactions depending on it will become invalid. This will make users unable to claim the refunds if there is.
-    Segregated Witness soft fork addresses this issue and therefore is an important upgrade to support Lightning Network.
+    A commitment transaction needs the hash of a funding transaction and if the hash of the funding transaction changes, transactions depending on it will become invalid. This will make users unable to claim the refunds if there are any.
+    The segregated Witness soft fork addresses this issue and is therefore an important upgrade to support Lightning Network.
```
giovannizotta: Segregated should be capitalized

glossary.asciidoc
```diff
@@ -488,13 +483,13 @@ transaction::
...
 
 transport layer::
-    In computer networking, the transport layer is a conceptual division of methods of a model of how computers talk to each other.
+    In computer networking, the transport layer is a conceptual model of how computers talk to each other.
```
giovannizotta: I'm not sure I love this cut

glossary.asciidoc
```diff
@@ -504,19 +499,18 @@ upstream payment::
     TBD.
 
 wallet::
-    Software that holds all your Bitcoin addresses and secret keys. Use it to send, receive, and store your bitcoin.
+    A walledt is a piece of software that holds Bitcoin addresses and secret keys. It is used to send, receive, and store bitcoin.
```
giovannizotta: Small typo in "walledt"

glossary.asciidoc
```diff
@@ -504,19 +499,18 @@ upstream payment::
...
     
 watchtower::
     Watchtowers are a security service on the Lightning network that monitor channels.
-    In the case that one of the channel partners goes offline or loses their backup, a watchtower keep their own backups and can restore their channel information.
-    They also monitor the Bitcoin blockchain and can submit a penalty transaction in the case that one of the partners tries to "cheat" by broadcasting an outdated state.
+    In case one of the channel partners goes offline or loses their backup, a watchtower keeps its own backups and can restore channel information.
```
giovannizotta: I think the old sentence was more correct

giovannizotta (commented):
Great work! Thanks @Zendo102 for the valuable revision. I added some comments in the request, maybe we could discuss a little about those. Also, another general thing that I'm not sure about is when to use "Bitcoin" over "bitcoin", but I wanted to comment about it here instead of doing a comment for every single one.

glossary.asciidoc
ghost: Agreed.  The style of the original came across a quite abript.  The meaning is the same.

glossary.asciidoc
ghost: My reasoning: A glossary should define and explain a term and not introduce new complexity.  What is "an onion" and a "hop of the onion"? What is SPHINX Mix Format?  I think it is good to have an example, but it should be understandable without having to find other terms in the Glossary.

glossary.asciidoc
ghost: Agreed.  It is the "more + particular" of the original that is the issue for me (using an adjective instead of an adverb).

glossary.asciidoc
ghost: Yes, sorry.  Glossaries should be neutral.  I just think "we" should be removed.

giovannizotta: Sure, I agree with the correction, I just wanted to signal the typo before it gets merged :)

glossary.asciidoc
giovannizotta: I agree, the SPHINX Mix Format sentence is kinda hard to understand without context. Should we keep the link to Wikipedia though? Or should we just remove it?

glossary.asciidoc
ghost: It is confusing.  What is used as the variables, the data (plural) or the ScriptSig (singular)?

My preference would be:
"ScriptSig or signature script is the data generated by a spender and it is almost always used as a condition to satisfy a pubkey script."

This keeps the focus on the ScriptSig and not on the data.

glossary.asciidoc
ghost: If you look earlier in the Glossary under HTLC you will find: "A Hashed TimeLock Contract or HTLC is a class of payments that use hashlocks and timelocks"
Also, see the link below which uses "Hashed TimeLock Contract" - note the camelcase capital "L" in TimeLock.
 https://academy.binance.com/en/glossary/hashed-timelock-contract

glossary.asciidoc
ghost: Agreed. I need to proofread more!

glossary.asciidoc
ghost: The problem is the concatenation of "divisions", "methods" and "model".  How can you have "methods of a model"?  I think the original wording is partially taken from Wikipedia.

How about: 
In computer networking, the transport layer is a conceptual division of the methods used by computers and applications to talk to each other.

glossary.asciidoc
ghost: Sorry! Yes!

glossary.asciidoc
ghost: The original appears grammatically incorrect.  Surely it must be either:

a watchtower keeps its  (singular)
watchtowers keep their (plural)

glossary.asciidoc
ghost: By the way, I'm not sure what I should do about typos like this.  Is there a way I can correct it in the pull request, or do I just leave it for someone else to sort out?  (I'm a GitHub newbie!)  Guidance welcome.

glossary.asciidoc
giovannizotta: Agreed, I like the sentence :)
By the way, I think it was an adaptation of the wikipedia page: https://en.wikipedia.org/wiki/Transport_layer

glossary.asciidoc
giovannizotta: Oh, I don't know then. As a reference for HTLC's definition i was using the routing.asciidoc chapter, but looking a bit on the web it seems like the two names are used interchangeably, so I'm really not sure about this one

glossary.asciidoc
giovannizotta: Now that I read the sentence again, I think both are correct, it just depends on how you interpret them. I could be wrong, but as far as I've understood, watchtowers keep backups of channel information of the two channel partners. Do you say that the watchtower keeps ITS own backup (intended as the backup kept by the watchtower) or do you say that the watchtower keeps THEIR backup (intended as the backup of the two channel partners) ?

glossary.asciidoc
giovannizotta: To be honest, I'm not a pro either :)
What I think you should do is: 
* on your local repository, use "git pull" to make sure you're in sync with the main repository
* do the changes that you want to make on your repository
* commit the changes as a new commit
* push the commit

This should appear as a new commit in this pull request, or at least I think so.

glossary.asciidoc
ghost: I'd remove the link.  It doesn't add authority, and the wording is different.  The readers of the book will know how to do a search!

glossary.asciidoc
ghost: Unfortunately, I don't know enough about how the technology works to make a judgement on that.

giovannizotta (approved):

glossary.asciidoc
renepickhardt: I am a little uncertain. I agree that for a glossary it is too long

glossary.asciidoc
```diff
@@ -158,7 +156,6 @@ digital signature::
 double-spending::
     Double-spending is the result of successfully spending some money more than once.
     Bitcoin protects against double-spending by verifying that each transaction added to the blockchain plays by the rules; this means checking that the inputs for the transaction have not previously already been spent.
-    The Revocable Sequence Maturity Contracts used to construct payment channels heavily attempt to double-spend bitcoin.
```
renepickhardt, line 161: While above I tend to agree that removing the example might be good in this particular case I tend towards keeping it. The RSMC and doublespending is in my opinion a very crucial part of the protocol

glossary.asciidoc
```diff
@@ -198,7 +195,7 @@ funding transaction::
     From the perspective of the Bitcoin network, the process of opening a channel by creating a RSMC is started by creating the funding transaction and finished by broadcasting it to the Bitcoin network and have it included in the blockchain.
     The value of the funding transaction is exactly the capacity of the payment channel.
     The output of the funding transaction is a 2-of-2 multisignature script (multisig) where each channel partner controls one key.
-    It is supposed to be spent by the commitment transactions or by the closing transaction.
+    It is spent by the commitment transactions or by the closing transaction.
```
renepickhardt: I think the word supposed is correct. It can happen that it is never spent or that ownsers of the channel define something complete different.

glossary.asciidoc
renepickhardt: so can we change this to the suggestion from @GiovanniZotta ?

glossary.asciidoc
```diff
@@ -317,13 +313,12 @@ onion routing::
     The encrypted data is transmitted through a series of network nodes called onion routers, each of which peels away a single layer, uncovering the data's next destination.
     When the final layer is decrypted, the message arrives at its destination.
     The sender remains anonymous because each intermediary knows only the location of the immediately preceding and following nodes.
-    With the SPHINX Mix Format, the final destination also remains anonymous as only the previous router could see it but does not know if they are routing it to the final node or just the next hop.
-    https://en.wikipedia.org/w/index.php?title=Onion_routing&oldid=870849217
+       https://en.wikipedia.org/w/index.php?title=Onion_routing&oldid=870849217
```
renepickhardt, line 319: Why did you remove that sentence?

glossary.asciidoc
```diff
@@ -362,45 +357,45 @@ peer::
...
 
 private channel:
     A channel not announced to the rest of the network. 
     Technically "private" is a misnomer as these channels can still be identified through routing hints and commitment transactions.
-    We suggest they are better described as "unannounced" channels.
+    They are better described as "unannounced" channels.
     With an unannounced channel, the channel partners can send and receive payments between each other as normal.
-    However the rest of the network will not be aware of it and so cannot typically use it to route payments.
-    As the number of and capacity of unanounced channels is unknown, the total public channel count and capacity only accounts for a portion of the total Lightning Network.
+    However, the rest of the network will not be aware of the channel and so cannot typically use it to route payments.
```
renepickhardt: cannot is actually not accurate. Better: 

so it will typically not be used by other nodes to route payments. (of course it can do it and it also might happen as some people might be aware of the `unannounced` channel)

glossary.asciidoc
renepickhardt: agree with the suggestion by @Zendo102

glossary.asciidoc
renepickhardt: good catch!

glossary.asciidoc
renepickhardt: I think we will just aim to be consistent in final copyediting

glossary.asciidoc
renepickhardt: I think `keeps its` can be used. as far as the backup / copy of channel state it's the watchtowers backup

renepickhardt (changes requested):
Excellent and critical work! thanks a lot!

giovannizotta:
Hey @Zendo102, just a quick poke. I think that if you resolve the changes proposed by @renepickhardt it's on its way to be merged!

ghost:
@GiovanniZotta Thanks for the poke!  I have a lot of experience editing and proofreading articles and books, but Github is new to me.  I thought I just had to suggest changes.   Anyway, this is just the motivation I need to delve into more chapters.

glossary.asciidoc
ghost: Yes.

glossary.asciidoc
ghost: Agreed.  Apologies for the slow response.

glossary.asciidoc
ghost: Ok, agreed.  (I'm learning about Lightning as I read, so it's difficult for me to judge the relative importance of information. )

glossary.asciidoc
ghost: Ok.  Again, my technical knowledge is lacking.  However, "supposed to be" has the idea that something should happen, but it didn't.  E.g. "I was supposed to go to work today."

How about changing it to:
It may be spent by the commitment transactions or by the closing transaction.

or
It may or may not be spent by the commitment transactions or by the closing transaction.

glossary.asciidoc
ghost: Here is my comment from before (I'm thinking like a moderate techie rather than a hard-core techie):

My reasoning: A glossary should define and explain a term and not introduce new complexity. What is "an onion" and a "hop of the onion"? What is SPHINX Mix Format? I think it is good to have an example, but it should be understandable without having to find other terms in the Glossary.

However, as I said in the thread above, I can't gauge the relative importance of information I don't know enough about.  If it is key for the understanding of the glossary by a techie, then keep it in.

glossary.asciidoc
ghost: Ok.  Agreed.

ghost:
I've added comments and resolved most of discussions that I think have come to a conclusion.

glossary.asciidoc
renepickhardt: I think your initial suggestion is actually acurate enough. following the protocol it it will eventually be spent by either the commitment or closing tx. we could add the world `eventually` but I think that might be too much detail.

renepickhardt (approved):
thanks a lot for this detailed ground work!

PR #676, opened 2021-04-02
docker builds images and not containers

RooSoft:
The command `docker build` creates images and not containers.


RooSoft:
Figuring out there are more references to containers that should be rather called images. If this pull request passes, I volunteer to also correct them.

GustavoRSSilva (approved):

PR #683, opened 2021-04-08
Remove redundant use of word

konstantinos-giannakakis:
Remove redundant use of word "this"


Roasbeef:
PR is dated, chapter has shifted significantly.

PR #689, opened 2021-04-16
Load wallet instead of creating a new one if it already existed

kevin-miller-0405:
This allows us to reuse the existing bitcoind container by doing a `docker start -i bitcoind`.

Otherwise, we'd have to `docker rm bitcoind` and then `docker run` it every time when we stop the container.


renepickhardt:
LGTM but please @aantonop decide (:

Roasbeef:
Latest version of the script incorporates this, but in a slightly different way.

PR #690, opened 2021-04-16
Update 05_node_operations.asciidoc

tecwesty:
Updated the first 'notes' section for clarity and subject consistency.


renepickhardt (changes requested):
unsure about this one. I think the proposed version sound a bit shorter but I think we have th `you` form also at other places, Please @aantonop decide

PR #694, opened 2021-04-17
minor glossary updates for consistency

senf42:
Italicized some references to other entries, changed a "Look at" to "See". Added periods when "See _X_" was a sentence ending a paragraph for an entry, removed periods when "See _X_" was the only part of the entry.


renepickhardt:
the dots at the line ends are not consistent anymore.

PR #713, opened 2021-05-20
WIP: Payment Channels Chapter (Channel Construction rewrite)

aantonop:
This is a work-in-progress chapter that will incorporate the draft from channel_construction.asciidoc in a new structure and outline.


renepickhardt:
while it is very short and reads good I am not sure if we want to talk about the establishment of a peer connection in this chapter or just move this out to the BOLT08 chapter?

renepickhardt:
Finished review up to making payments and moving state. Included a few additional / further thoughts and made some tiny changes. Looks very good so far!

Hope I can finish the state part tomorrow.

renepickhardt:
also tagging issues #722, #723, #724 and #725 as they came up in the process and might be relevant

renepickhardt:
Need to to still finish everything from advancing channel state. 

That being said. I think it might make sense to drop that part here as we have channel operation as an own chapter and we could just explain here on a very high level that alice and bob need to somehow move the state forward and that for simplicity / reusability in the routing case this is also done via HLTCs 

also there is one thing where I am currently confused in the scripts with the revocation key. I believe the non time locked spending condition of to_self should require two keys. One from the channel partner and the revocation pubkey. If it was just the revocation pubkey the node pushing the commitment transaction could just use that key to claim the outputs without time lock. but that person should never be able to use the revocation pubkey thouth it certainly knows it as it needs to be shared in `revoke_and_ack` when revoking the state. 

That being said we should note (and I think I did that in channel operation) that the revocation pubkey changes with every balance / state that was negotiated otherwise sharing it once would work for every future commitment transaction (and in particular the most recent one that was not to be revoked yet). in that sense we have not discussed the idea that the balances are a sequence of commitment transactions (though that was certainly touched and is even depicted in the pictures)

While that sounds like many points I have to say that the chapter finally reads very well. While it seems like it took us forever to finally arrive here I think after reading it now that the time was very well spent and invested! Really good team effort and great rewrite @aantonop

PR #730, opened 2021-07-05
Security and Privacy - Private Channels

ImranLorgat:
Clarified and expanded section on Private Channels

### DIFF
--- a/security_privacy_ln.asciidoc
+++ b/security_privacy_ln.asciidoc
@@ -565,18 +565,30 @@ We're still in the early stages of the Lightning Network.
 Many of the concerns listed in this chapter are likely to be addressed as it matures and grows.
 In the meantime, there are some measures that you can take to guard your node against malicious users; something as simple as updating the default parameters that your node runs with can go a long way in hardening your node.
 
-== Private channels
-If you intend to use the Lightning Network to send and receive funds between nodes and wallets you control, there is little need to open public channels to the rest of the network.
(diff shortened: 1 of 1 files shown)


renepickhardt (changes requested):
I am actually not sure if we should / want to call them private channels. it has been pointed out several times that they are much more `unannounced` and I think this is where the terminology among lightning devs is heading.

ImranLorgat:
Noted. Updated this section and expanded a little with privacy caveats.

Not sure what happened to the chapter. It doesn't seem to be in the repo anymore.

renepickhardt:
It's been moved to contrib folder. I think the idea was to have everything that is still WIP in contrib. Can you confirm @aantonop ?

ImranLorgat:
I see it's there. Link below. Are you able to merge the changes into the moved file, or do I need to open another pull request?

https://github.com/lnbook/lnbook/blob/develop/contrib/security_privacy_ln.asciidoc#private-channels

ImranLorgat:
@aantonop @renepickhardt I'm following up on this

I see Chapter 13 is marked as "published as an early draft". The section on private channels still refers to them as "private" instead of as "unannounced" and is the original brief section. Have the changes in this pull request been rejected?

https://github.com/lnbook/lnbook/blob/develop/13_security_privacy_ln.asciidoc#private-channels

Roasbeef:
Can you rebase this? The text has shifted slightly, and it's also now chapter 15.

PR #733, opened 2021-07-08
changed combined to combine (typo)

willtech3:


renepickhardt (changes requested):
uncertain about grammar. in my oppinion both versions are incorrect. 

I would write: 

`the unlocking script combined with the locking script would produce`

but I'd rather have @aantonop have a look

willtech3:
I think Rene's suggested text is better.

renepickhardt (approved):
Thx!

PR #736, opened 2021-07-09
Onion routing chapter release prep

aantonop:


Roasbeef:
I found it easier to do edits/review by branching off your current version and adding commits on top of that: https://github.com/aantonop/lnbook/compare/onion_routing_release...Roasbeef:onion-chapter-edits?expand=1 

Shooting to finish getting through the entire thing tomorrow, but from here we can either have me make a PR to your personal branch, or cherry-pick my commits in from there.

Roasbeef:
Alright, I've finished an initial pass all the way through. Excellent work on this chapter @aantonop! Really wish I had such a cogent explanation of the Sphinx packet construction+processing back in 2015 when I spent many nights/weekends at the desk in my dorm room trying to figure out how tf things actually worked 😅. The diagrams are amazing IMO, and really help to guide the user through the entire process. By the time they get to the later hops in the route it really just becomes "rinse and repeat". 

During editing I added a section on replay protection+detection since that wasn't covered and is a very important component for the security of the scheme. I also slightly elaborated on the way that the session keys are re-randomized at each hop so the sender doesn't need to blow up the packet size by including a session key for each hop. As noted in the TODO section, what's still missing here is a precise description of how the blinding factors are used to deterministically randomize for session key each hop. 

While reading I spotted one minor error (only really mentioned at the start): there isn't another "pad" key used to generate the fillers, instead just the rho key is used for this purpose. 

The other (more fundamental) error I found is the way the generation and verification of the HMACs is described:
  * An HMAC doesn't need to be computed for the final hop, it's actually all zeros. 
  * The HMAC is always computed over the "latest" _encrypted_ packet, not any of its more plaintext versions. 
  * Each hop only ever verifies the HMAC once, the only other operation related to the HMAC they perform is taking the emplaced "inner" HMAC, and using it as the "outer" HMAC for the next hop. 

I've attempted to fix the HMAC processing descriptions in line where applicable. However the diagrams that describe the HAMC computation will need to be _redone_ to reflect the fact that they're computed over the encrypted version of the packets (encrypt-then-mac).

PR #740, opened 2021-07-12
ch-07: technical edits to payment channels chapter 

Roasbeef:
A first pass of edits focused on technical accuracy through the current payment channels chapter. I've added a few important things not explicitly mentioned such as the existence of the upfront shutdown address, and the role of state hints during breach retribution. As usual, all edits are contained in individual commits so we can easily throw out certain edits or modify them.


07_payment_channels.asciidoc
```diff
@@ -68,7 +68,7 @@ Let's assume that Alice wants to construct a payment channel allowing her to con
 Every node on the Lightning Network is identified by a _node public key_. The public key uniquely identifies the specific node and is usually presented as a hexadecimal encoding. For example, Rene Pickhardt currently runs a Lightning Node (+ln.rene-pickhardt.de+) that is identified by the following node public key:
 
 ----
-03efccf2c383d7bf340da9a3f02e2c23104a0e4fe8ac1a880c8e2dc92fbdacd9df
+02a1cebfacb2674143b5ad0df3c22c609e935f7bc0ebe801f37b8e9023d45ea7b8
```
renepickhardt, line 71: I am fine with having my public node id there it's connected to my identity anyway

07_payment_channels.asciidoc
```diff
@@ -259,6 +259,9 @@ Alice's node constructs a multisignature script as shown in <<A_B_multisig>> bel
 2 <Alice_funding_pubkey> <Bob_funding_pubkey> 2 CHECKMULTISIG
 ----
 
+Note that in practice, the funding keys are deterministically _sorted_ (using lexicographical order of the serialized compressed form of the public keys) before being placed in the witness script (. By agreeing to this sorted order ahead of time, we ensure both parties will construct an identical funding transaction output, which is signed by the commitment transaction signature exchanged.
```
renepickhardt, line 262: maybe as an info box? (but I think very useful to mention this!)

renepickhardt (changes requested):
We are still missing the part that the revocation key is basically only a partial key through linearity the homomorphism of going from private keys aka the elements in the additive group of Z/pZ to the Elliptic curve whose elements are the public keys. I think we really have to add that as I know it is confusing for many to have a revokation key when in reality it is only a partial key

Roasbeef:
>  I think we really have to add that as I know it is confusing for many to have a revokation key when in reality it is only a partial key

Agreed, working on a follow up PR inspired by our conversation to explain why it would be insecure to just have a pubkey there that's 100% under control of the opposite party (they can just breach then take that path), then walk the reader through why the construction used today works in practice.

PR #746, opened 2021-07-15
Remove duplicated info about autopilot

jamesdorfman:
This PR removes a paragraph which describes autopilot, because a nearly-identical one appears a bit later in the chapter, in the **autopilot** section.

Both sections are virtually word-for-word the exact same.  It does not make sense to include the same content twice, and so close together.


renepickhardt (approved):
fully agree! very good catch. easy to have such leftovers in the text while reorganizing them and it is great if someone catches them!

PR #749, opened 2021-07-20
"A lot of electricity" has political connotations

Camilojdl:
Instead of "Miners have to use a lot of electricity doing "work" that is embedded as a "proof""

I propose: "Miners have to use electricity and dedicate hardware doing "work" that is embedded as a "proof""


renepickhardt (approved):
agree we either quantify it or we just state that electricity is used for computation

PR #750, opened 2021-07-21
WIP: Path Finding / Payment Delivery Chapter - Release prep

aantonop:


path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
+[[path_finding]]
+== Path Finding and Payment Delivery
+
+Payment delivery on the Lightning Network depends on finding a path from the sender to the recipient, a process called _path finding_.
```
renepickhardt: Suggestion for the intro: 

Delivering a payment is a classical transportation problem. You want to transport some Satoshis from your node / your channels along one or several paths to a recipient. 

As the Network is heavily connected many candidate paths exist and the main question that we address in this chapter is which one(s) to choose? 

The problem is in particular difficult for us since we don't know if remote channels can actually fulfill the routing requests from out onions as we do not know their local balance. Which means we do have uncertainty. This is why we often refer to this techniques and approaches in the chapter as path finding since we literally are trying to find paths with enough liquidity through the network that allow the transportation of the satoshi amount

path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+[[lngraph]]
+image::images/LNGraphJuly2021.png[]
+
+If the sender and recipient are connected to other well-connected nodes and have at least one channel with adequate capacity - there will be thousands of paths. The problem becomes: choosing the _best_ path that will succeed in payment delivery, out of a list of thousands of possible paths.
+
+==== Selecting the best path
+
+To select the "best" path, we have to first define what we mean by "best". There may be many different criteria such as:
+
+* Paths with enough capacity. Obviously if a path doesn't have enough capacity to route our payment, then it is not a suitable path
```
renepickhardt: here I would use the word `balance` or `liquidity` of course this is constrained by the `capacity` of the channel. And I woul already emphasize the uncertainty about it which can (similar to the next two points `fees` and `timelocks` be interpreted as a cost)

path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+* Paths with short timelocks. We may want to avoid locking our funds for too long and therefore select paths with shorter timelocks.
+
+All of these criteria may be desirable to some extent and selecting paths that are favorable across many dimensions is not an easy task. Optimization problems like this may be too complex to solve for the "best" solution, but often can be solved for some approximation of the optimal. Which is good news, because otherwise path finding would be an intractable problem.
+
+=== Fundamentals of path finding
+
+The problem of path finding in the Lightning Network falls under a general category of _graph theory_ in mathematics and the more specific category of _graph traversal_ in computer science.
+
+A network such as the Lightning Network can be represented as a mathematical construct called a _graph_, where _nodes_ are connected to each other by _edges_ (equivalent to the payment channels). The Lightning Network forms a _directed graph_ because the nodes are linked _asymmetrically_, since the channel capacity is split between the two channel partners and the payment capacity is different in each direction.
```
renepickhardt: `channel capaciy` might be confusing here as it is meant differently than the `capacity` of the channel announced over gossip. 

in the glossary I think we deine the term (local) balance or inbound / outbound capacity. I thin for understanding it is very crucial to basically say:  Look the ChannelGraph from gossip has capacities and is undirected but in reality this leads to a directed `balance graph` where the balance values on the edges A---->B and B---->A between two peers A & B on the balance graph add up to the capacity value of the edge A----B (note: undirected!) on the capacity graph.

path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+With only partial information about the network topology available this is a real challenge and active research is still being conducted into optimizing this part of the Lightning Network implementations.
+
+The fact that the path finding problem in the Lightning Network is not "fully solved" is a major point of criticism of the technology. Some critics go as far as stating that the problem is intractable and therefore the Lightning Network can't work at any significant scale. Obviously, we disagree.
```
renepickhardt: I am uncertain here: We might not even want to address the fud. But if we do we could just point to the recent paper

path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+
+A network such as the Lightning Network can be represented as a mathematical construct called a _graph_, where _nodes_ are connected to each other by _edges_ (equivalent to the payment channels). The Lightning Network forms a _directed graph_ because the nodes are linked _asymmetrically_, since the channel capacity is split between the two channel partners and the payment capacity is different in each direction.
+
+==== Uncertainty of balances
+
+If we knew the exact channel balances of every channel, we could easily compute one or more payment paths using any of the standard path finding algorithms taught in good computer science programs. But we don't know the channel balances, we only know the aggregate channel capacity, which is advertised by nodes in channel announcements. In order for a payment to succeed, there must be adequate balance on the sending side of the channel. If we don't know how the capacity is distributed between the channel partners, we don't know if there is enough balance in the direction we are trying to send the payment.
```
renepickhardt: if the uncertainty is expressed properly we can aply the same algorithms as before. 

for single paths dijkstra with `-log(P(X>= amt))` as edges weights works just fine and the proper generalization is `min cost flows` with the same cost function. 

the same argument that you present holds. if we did not have uncertainty we could solve single path with disktra and the fee function as weights or min cost flow for cheapest multi part payment with fee funciton as weight. 

in reality we do of course a combination of costs from uncertainty and fees and in practice many implementations just do single parts (even in the mpp case but I am certain that this will change)

path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+With only partial information about the network topology available this is a real challenge and active research is still being conducted into optimizing this part of the Lightning Network implementations.
+
+The fact that the path finding problem in the Lightning Network is not "fully solved" is a major point of criticism of the technology. Some critics go as far as stating that the problem is intractable and therefore the Lightning Network can't work at any significant scale. Obviously, we disagree.
+
+==== Path finding complexity
+
+Finding a path through a graph is a problem modern computers can solve rather efficiently.
+Developers mainly choose breadth-first search if the edges are all of equal weight.
+In cases where the edges are not of equal weight, the Dijkstra Algorithm is used.
+In our case the weights of the edges can represent the routing fees.
```
renepickhardt: see above comment: yes they `can` represent routing fees. the main problem is if you want to deliver 99k sats and there is a really cheap 100k sats channel those algorithms will try that. 

However expressing the uncertainty is as easy as expressing the routing fees by using `-log(P(X>= amt)`

I think just because we didn't look at the problem for along time we should not be afraid to just name it in this way

path_finding.asciidoc
```diff
@@ -0,0 +1,76 @@
...
+This turns our easy theoretical computer science problem into a rather complex real-world problem.
+We now have to solve a path finding problem with only partial knowledge.
+For example, we suspect which edges might be able to forward a payment because their capacity seems big enough.
+But we can't be certain unless we try it out or ask the channel owners directly.
+Even if we were able to ask the channel owners directly, their balance might change by the time we have asked others, computed a path, constructed an onion and send it along.
+Not only do we have limited information but the information we have is highly dynamic and might change at any point in time without our knowledge.
+
+==== Keeping it simple
+
+The simplest path finding strategy currently implemented in Lightning nodes is to probe paths (by attempting to deliver a payment) until a path is found that has enough liquidity to deliver the payment.
```
renepickhardt: That strategy is actually always applied! the only question is: what to optimize for?

It is btw funny: The current cost function used by implementations which mainly optimizes for fees works actually only well for small amounts as you noted (however for small amounts the fees should not be too much of a problem as the amount is small)

renepickhardt (changes requested):
I added a few comments that might help to find our general perspective to the problem

renepickhardt:
Just summarizing here what we have discussed: 
 
1. there are three terms which we might want to describe: 
 * **Capacity**: (as announced on the gossip)
 * **Balance**: (as defined in the commitment transaction's to_local output (in flight htlcs do not account for balance on either side yet)
 * **Liquidity** = Balance - channel reserve (if and only if `balance - channel reserve > min_htlc_size`
2. Min-cost flow for MPP is the proper generalization of single part (e.g. bfs (just connectivity no optimization for fees or anything) dijsktra or A* or what ever algorithm / heuristic to use
3. Uncertainty is always a range in which the **liquidity** is (worst case: `[channel reserve, capacity - channel reserve]` (ignoring reserve it woult be `[0, capacity]`) best case `[exact_value, exact_value]`
4. Uncertainty is the main reason why we need trial and error loop and it is the main limiting factor for sending larger amounts quickly, realiably. 
5. even if no uncertainty we would still want to solve an optimization problem in Single path and MPP case (with cost function maybe fees or cltv delta)
6. Taking uncertainty into account of the cost function as (`-log(P(X>= amt)`) has makes sense (similar as fees make sense) 
7. there are three different networks involved
 * the **channel graph** which we observe from gossip it is (when ignoring fees) undirected but with fees it is already directed. Probably best to not talk about fees or other cost-functions here
  * the **balance graph** or **liquidity graph** here the edges are limited by the actual liquidity that could be send on the network. This graph depicts how the actual lightning network looks at some point in time. But large fractions of this graph are unknown to many users most of the time. (Note liquidity is always between `[0, capacity -2*channel_reserve]`)
  * the **uncertainty network** which defines ranges of uncertainty in which range the liquidity is and which can be learned from onion errors during the payment loop or by probing before.

I hope I didn't miss anything.

path_finding.asciidoc
```diff
@@ -58,6 +58,15 @@ If we knew the exact channel balances of every channel, we could compute one or
...
+
+----
+min(liquidity) = channel_reserve
+max(liquidity) = capacity - channel_reserve
+----
+
+Our channel liquidity uncertainty range is the range between the minimum and maximum possible liquidity. This is unknown to the network, except the two channel partners. However, as we will see we can use failed HTLCs returned from our payment attempts to update our liquidity estimate and reduce uncertainty. If for example we get an HTLC failure code that tells us that a channel cannot fulfill an HTLC that is smaller than our estimate for maximum liquidity, that means the maximum liquidity can be updated to the amount of the failed HTLC. In simpler terms, if we think the liquidity can handle an HTLC of N satoshis and we find out it fails to deliver M satoshis (where M is smaller), then we can update our estimate to M-1 as the upper bound. We tried to find the ceiling and bumped against it, so it's lower than we thought!
+
```
renepickhardt: We can also update the minimum liquidity estimate on all upstream channels that the htlc traveled before it hit `Temporary_channel_failure` which is the message coming if an htlc cannot be passed to a downstream channel (unfortunately this message also seems to come in other non liquidity related issues)

renepickhardt (approved):
I ACK what you wrote but point out that we also learn from the successful channels until the erring one.

path_finding.asciidoc
```diff
@@ -12,11 +12,7 @@ But, as the Lightning Network has grown explosively, the path finding problem's
 [[lngraph]]
 image::images/LNGraphJuly2021.png[]
 
-If the sender and recipient are connected to other well-connected nodes and have at least one channel with adequate capacity - there will be thousands of paths. The problem becomes: choosing the _best_ path that will succeed in payment delivery, out of a list of thousands of possible paths.
-
-Delivering a payment can be described as a _transportation problem_. You want to transport some satoshis from your node via your channels and along one or several paths to a recipient.
-
-The problem is difficult because we don't know if remote channels can actually fulfill the routing requests as we do not know their local balance. Which means we have uncertainty. This is why we refer to path finding in this chaptersince we are trying to find paths with _enough liquidity_ through the network that allow the transportation of the satoshis.
+If the sender and recipient are connected to other well-connected nodes and have at least one channel with adequate capacity - there will be thousands of paths. The problem becomes: selecting the _best_ path that will succeed in payment delivery, out of a list of thousands of possible paths.
```
renepickhardt: I would use plural because in general we look for a collection of paths. Only for very small amounts this will result in a single path

renepickhardt (changes requested):
Would change path to plural here

path_finding.asciidoc
```diff
@@ -7,11 +7,16 @@ Payment delivery on the Lightning Network depends on finding a path from the sen
...
 
-But, as the Lightning Network has grown explosively, the path finding problem's nature has shifted. In mid-2021, as we finish this book, the Lightning Network consists of 20,000 nodes connected by at least 55,000 public channels with an aggregate capacity of almost 2,000 BTC. A node has on average 8.8 channels, while the top 10 most connected nodes have between 400 and 2000 channels _each_. A visualizationfootnote:[Produced by explorer.acinq.co from published channel updates] of just a small subset of the Lightning Network channel graph (as of July 2021) is shown in <<lngraph>>:
+But, as the Lightning Network has grown explosively, the path finding problem's nature has shifted. In mid-2021, as we finish this book, the Lightning Network consists of 20,000 nodes connected by at least 55,000 public channels with an aggregate capacity of almost 2,000 BTC. A node has on average 8.8 channels, while the top 10 most connected nodes have between 400 and 2000 channels _each_. A visualization of just a small subset of the Lightning Network channel graph (as of July 2021) is shown in <<lngraph>>:
```
renepickhardt: In network science the average node degree is usually a pretty meaning less metric as that distribution follows usually a zip'f style power law. I understand that many people might have issues with a power law coefficient thus I suggest that we just depict a histogram of the degree distribution. I am happy to provide one and depict a bit of text (guess the cummulative distribution function will as usual make most sense)

path_finding.asciidoc
```diff
@@ -122,3 +127,21 @@ These three activities can be repeated in a _payment round_ if we use the failur
...
+
+node_announcement:: This contains the information about a node on the Lightning Network, such as its node ID (public key), network address (e.g. IPv4/6 or Tor), capabilities/features etc.
+
+channel_announcement:: This contains the capacity and channel ID of a public (announced) channel between two nodes and proof of the channel's existence and ownership.
+
+channel_update:: This contains a node's fee and timelock (CLTV) expectations for routing an outgoing (from that node's perspective) payment over a specified channel.
+
+In terms of a mathematical graph, the node_announcement is the information needed to create the nodes or _vertices_ of the graph. The channel_announcement allows us to create the _edges_ of the graph representing the payment channels. Since each direction of the payment channel has its own balance, we create a directed graph. The channel_update allows us to incorporate fees and timelocks to set the _cost_ or _weight_ of the graph edges.
+
+Depending on the algorithm we will use for path finding, we may establish a number of different cost functions for the edges of the graph.
```
renepickhardt: I think in all cases we use the same algorithm it really is just the cost function that changes.

Proposal: 

"Depending on our optimization goal for transporting the funds, we may establish a number of different cost functions (weights) for the edges of the graph"

we could append the sentence with "to be used in the path finding or min cost flow algorithm" (however I think that would be too much detail for now).

path_finding.asciidoc
```diff
@@ -122,3 +127,21 @@ These three activities can be repeated in a _payment round_ if we use the failur
 In the next sections we will look at each of these steps in more detail, as well as more advanced payment strategies.
 
 === Channel graph construction
+
+In <<gossip>> we explain the three main messages that nodes "gossip": +node_announcement+, +channel_announcement+, and +channel_update+. These three messages allow any node to gradually construct a "map" of the Lightning Network in the form of a _channel graph_. Each of these messages provides a critical piece of information for the channel graph:
+
+node_announcement:: This contains the information about a node on the Lightning Network, such as its node ID (public key), network address (e.g. IPv4/6 or Tor), capabilities/features etc.
+
+channel_announcement:: This contains the capacity and channel ID of a public (announced) channel between two nodes and proof of the channel's existence and ownership.
```
renepickhardt: I assume this is already incorperating gossip? In which case I would explain how the short channel id has the standardized format `height`x`index`x`output`

path_finding.asciidoc
```diff
@@ -122,3 +127,21 @@ These three activities can be repeated in a _payment round_ if we use the failur
...
+
+Depending on the algorithm we will use for path finding, we may establish a number of different cost functions for the edges of the graph.
+
+For now, let's ignore the cost function and simply establish a channel graph showing nodes and channels, using the node_announcement and channel_announcement messages. In this section we will use a network of six nodes connected by seven channels. One node (S) is our sender, one node (R) is our recipient. Nodes A, B, X, and Y are nodes that announced themselves with a node_announcements. We've also received seven channel_announcement messages with channel capacities, allowing us to construct a basic "map" of the network, shown in <<channel_graph_1>>, below:
+
+[[channel_graph_1]]
+.The channel graph
+image::images/channel_graph_1.png[]
```
renepickhardt: Just noting here that you have multiplied all capacities with 100k which will destroy part of the original reason for the example showing that the min cost flow of 1 unit (which included the X-B edge) does not have to be a subflow of the min cost flow for 2 units which did not include the X-B edge) I guess we will not go into that much detail but also all probability computations (if we include some) will become nasty numbers (even if we multiplay sent amounts also by 100k) as the numbers are always `(cap+1 - flow)/(cap + 1)` when we stay with 1,2,4,7,9 we have easy fractions

renepickhardt: so for the sake of long term understandability I might suggest to stick with the more artificial example

path_finding.asciidoc
renepickhardt: also I may add that we could exchange the word explosively to exponentially (though this is note true for some time windows)

renepickhardt (changes requested):
if you agree with my comments about average node degree I can create a histogram and some text around it (I will focus hard on not just describing it in scientific words but boiling it down (without opening pandoras box))

path_finding.asciidoc
```diff
@@ -140,8 +140,43 @@ In terms of a mathematical graph, the node_announcement is the information neede
...
+In this example, Selena is constructing a channel graph. Selena will use her channel graph to find a path to send a payment to Rashid. This is _Selena's channel graph_. There is no such thing as *the* channel graph, there is only ever *a channel graph* and it is always from the perspective of the node that has constructed it (see <<map_territory_relation>>).
+
+[[map_territory_relation]]
+.The map-territory relation
+****
+From Wikipedia's https://en.wikipedia.org/wiki/Map%E2%80%93territory_relation[_Map Territory Relation_], "The map–territory relation describes the relationship between an object and a representation of that object, as in the relation between a geographical territory and a map of it."
+
+The map territory relation is best illustrated in "Sylvie and Bruno Concluded", a short story by Lewis Carroll which describes a fictional map that is 1:1 scale of the territory it maps, therefore having perfect accuracy but becoming completely useless as it would cover the entire territory if unfolded.
+
+What does this mean for the Lightning Network? LN is the territory, and a channel graph is a map of that territory:
```
renepickhardt: I love this! even when ignoring the uncertainty of the liquidity but just taking the channel graph from gossip this is already true!

path_finding.asciidoc
```diff
@@ -140,8 +140,43 @@ In terms of a mathematical graph, the node_announcement is the information neede
...
+
+While we could imagine a theoretical (Platonic ideal) channel graph that represents the complete, up-to-date map of the Lightning Network, such a map is simply the Lightning Nwtwork itself. Each node has its own channel graph which is constructed from announcements and is necessarily incomplete, incorrect, and out-of-date!
+
+The map can never completely and accurately describe the territory.
+****
+
+Selena listens to node_announcement messages and discovers 4 other nodes (in addition to Rashid, the intended recipient). The resulting graph represents a network of six nodes: (S)elena and (R)ashid are the (S)ender and (R)ecipient respectively; (A)lice, (B)ob, (X)avier and (Y)an are intermediary nodes. Selena's initial graph is just a list of nodes, shown in <<channel_graph_nodes>>:
```
renepickhardt: names! We should have done this in the paper! Amazing!

path_finding.asciidoc
```diff
@@ -140,8 +140,43 @@ In terms of a mathematical graph, the node_announcement is the information neede
...
 [[channel_graph_1]]
 .The channel graph
 image::images/channel_graph_1.png[]
+
+
+==== Uncertainty in the channel graph
+
+As you can see from <<channel_graph_1>>, Selena does not know any of the balances of the channels. Her initial channel graph contains the highest level of uncertainty.
+
+But wait: Selena does know *some* channel balances! She knows the balances of the channels that her own node has connected with other nodes. While this does not seem like much, it in fact very important information for constructing a path - Selena knows the actual liquidity of her own channels. Let's update the channel graph to show this information. We will use a "?" symbol to represent the unknown balances, as shown in <<channel_graph_2>>:
```
renepickhardt: As mentioned we switched from ? to intervals `0<= x <= capacity` or `b<= x<=b` which leads to `x = b` if we never talk about min balance / max balance or htlc in flight I think the question mark notation is fine!

renepickhardt (approved):
ACK though I left a few comments

path_finding.asciidoc
```diff
@@ -3,6 +3,22 @@
...
+
+Path finding, path selection, multi-path payments (MPP) and the payment attempt trial & error loop occupy the majority of the "Payment Layer" at the top of the protocol suite.
+
+These components are highlighted by a double outline in the protocol suite, shown in <<LN_protocol_pathfinding_highlight>>:
+
+[[LN_protocol_routing_highlight]]
+.The Lightning Network Protocol Suite
+image::images/LN_protocol_routing_highlight.png["The Lightning Network Protocol Suite"]
+
+==== Where is the BOLT?
```
renepickhardt: very important remark!

path_finding.asciidoc
```diff
@@ -3,6 +3,22 @@
 
 Payment delivery on the Lightning Network depends on finding a path from the sender to the recipient, a process called _path finding_.
 
+=== Path finding in the Lightning protocol suite
```
renepickhardt: I am still not sure if we might want to switch to something like `Finding Liquidity` or `liquidity finding` by the end of the day this issue is way more a transportation problem (we want to transport an amount `amt` from `S` to `R`)

path_finding.asciidoc
```diff
@@ -180,3 +198,15 @@ But wait: Selena does know *some* channel balances! She knows the balances of th
 [[channel_graph_2]]
 .Channel graph with known and unknown balances
 image::images/channel_graph_2.png[]
+
+==== Fees and other channel metrics
+
+==== Finding candidate paths
+
+==== Selecting a path, ordering of paths
```
renepickhardt: Concept NACK see longer comment for the review

renepickhardt (changes requested):
basically as mentioned in the very fist comment about the section name we solve a transportation problem through a network with costs at the edges.  Even if we assume full knowledge of balance values and when just optimizing for fees it is very unlikely to find the cheapest way to transport an amount  `amt` by just doing single paths and knowing how to do them best. (unless we are in the case where the optimal solution is a single path - which can happen for small amounts)

However when looking at this as a flow problem and starting from MPP we will see that in many instances of payment pairs and amounts the optimal payment flow will actually just be a single path but there is just this moment where splitting starts to make sense and strategies that start from a single path will fail at that moment / amount. 

This will lead towards solutions that are acutally far from optimal in both cases when just optimizing for fees or when trying to maximize the probability. 

Thus I think we should in this book - which will probably be an entry point for many people to learn about lightning - take on the glasses of operation research / logistics and transportation problems and thus start from there. 

Actually there is a mathematical theorem that flows can always be dissected into circles and paths (as circles don't do anything in our case we can assume a dissection into paths) and again in many cases the optimal flow will be just a single path.

path_finding.asciidoc
```diff
@@ -199,19 +210,65 @@ But wait: Selena does know *some* channel balances! She knows the balances of th
...
 
+While the "?" symbol seems ominous, a lack of certainty is not the same as complete ignorance. We can _quantify_ the uncertainty and _reduce_ it by updating the graph with the successful/failed HTLCs we attempt.
+
+Uncertainty can be quantified, because we know the maximum and minimum possible liquidity and can calculate probabilities for smaller (more precise) ranges.
+
+Once we attempt to send an HTLC we can learn more about channel balances: if we succeed, then the balance was *at least* sufficient to transport the specific amount. Meanwhile if we get a "temporary channel failure" error, the most likely reason is a lack of liquidity for the specific amount.
+
+[TIP]
+====
+You may be thinking "What's the point of learning from a successful HTLC?" After all, if it succeeded we're "done". But consider that we may be sending one part of a multi-part payment. We also may be sending other single-part payments within a short time. Anything we learn about liquidity is useful for the next attempt!
```
renepickhardt: Even during a single part payment in the trial and error loop it is useful as an onion might have failed close to the receiving node. one can now start path finding from all upstream nodes for the next attempt. 

of course somewhere we should mention that the knowledge is just temporary

path_finding.asciidoc
```diff
@@ -199,19 +210,65 @@ But wait: Selena does know *some* channel balances! She knows the balances of th
...
 
 ==== Finding candidate paths
 
-==== Selecting a path, ordering of paths
+Finding a suitable path through a directed graph like this is a well-studied computer science problem (known broadly as the "Shortest Path problem"), which can be solved by a variety of algorithms depending on the desired optimization and resource constraints.
+
+The most famous algorithm solving this problem was invented by Dutch mathematician E. W. Dijkstra in 1956, known simply as https://en.wikipedia.org/wiki/Dijkstra's_algorithm[_Dijkstra's Algorithm_]. In addition to the original Dijkstra algorithm, there are many variations and optimizations, such as https://en.wikipedia.org/wiki/A*_search_algorithm[A* ("A-star")], which is a heuristic-based algorithm.
+
+As mentioned previously, the "search" must be applied _backwards_ to account for fees that are accumulated from recipient to sender. Thus, Dijkstra, A* or some other algorithm would search for a path from the recipient to the sender, using fees, estimated liquidity, timelock delta (or some combination) as a cost function for each hop.
```
renepickhardt: I would at least make a stetment that another technique is to find flows and flows can always be disected into paths. 

In general (as mentioned) min-cost flows are the proper generalization of single min cost path problem (that dijkstra solves) so in particular if a single path is the result of a min-cost flow computation the resullt will be the same as disjktra (assuming the same cost function (depending on what ever feature we wish)

path_finding.asciidoc
```diff
@@ -199,19 +210,65 @@ But wait: Selena does know *some* channel balances! She knows the balances of th
...
 
 ==== Payment delivery (Trial-and-error loop)
 
-==== Updating the channel graph
+Once Selena has a selected path, her node starts the trial-and-error loop, by constructing the HTLCs, building the onion and attempting delivery of the payment. For each attempt, there are three possible outcomes:
+
+- A successful result (+update_fulfill_htlc+)
+- An error (+update_fail_htlc+)
+- A "stuck" payment with no response (neither success, nor failure)
```
renepickhardt: here we should differentiate as there are two forms of stuck payments (which we for now cannot differentiate on a protocol level) when we do MPP a path can be arrived at the recipient and all the HTLCs are locked in but the path will not settle yet as other onions need to arrive to achieve the full amount. That is what I would call an `expected` stuck payment or a locked htlc. then of course ther are the stuck payments (which I guess you had in mind when you wrote this) that are due to poor / hostile behaviour of routing nodes that just delay forwarding the htlcs or errors.

renepickhardt (approved):
concept ack. great progress. 

With regard to my previous comment I would consider not  to make this artificial distinction between single path payments and MPP in both cases we solve a transportation problem and if the amounts are sufficiently small (depending on the payment pair) the flow will be a single path anyway.

path_finding.asciidoc
```diff
@@ -269,6 +286,55 @@ If the payment fails, then it can be re-tried via a different path by updating t
...
+
+Fortunately, Selena receives an +update_fulfill_htlc+ message from Alice, indicating that the payment was successful, as shown in <<path_4_success>>:
+
+[[path_4_success]]
+.Path 4 attempt succeeds
+image::images/path_4_success.png[]
+
+===== Learning from success
+
+Selena has also learnt a lot from this successful payment. She now knows that all the channels on the path S->A->B->X->Y->R  had enough liquidity to deliver the payment. Furthermore, she now knows that each of these channels has moved the HTLC amount (1m + fees) to the other end of the channel. This allows Selena to recalculate the range of liquidity on the receiving side of all the channels in that path, replacing the minimum liquidity with 1m+fees.
```
renepickhardt: But also on the sending! Before it was [0,c] now it is [0,c-1m] (as 1 m is on the other side)

renepickhardt (changes requested):
Lgtm

path_finding.asciidoc
```diff
@@ -269,6 +286,55 @@ If the payment fails, then it can be re-tried via a different path by updating t
...
+
+Selena attempts the first path (S->A->B->R). She constructs the onion and sends it, but receives a failure code from Bob's node. Bob reports back a +temporary channel failure+ with a +channel_update+ identifying the channel B->R as the one that can't deliver. This attempt is shown in <<path_1_fail>>:
+
+[[path_1_fail]]
+.Path 1 attempt fails
+image::images/path_1_fail.png[]
+
+===== Learning from failure
+
+From this failure code, Selena will deduce that Bob doesn't have enough liquidity to deliver the payment to Rashid on that channel. Importantly, this failure narrows the uncertainty of the liquidity of that channel! Previously, Selena's node assumed that the liquidity on Bob's side of the channel was somewhere in the range (0, 4m). Now, she can assume that the liquidity is in the range (0, 999999). Similarly, Selena can now assume that the liquidity of that channel on Rashid's side is in the range (1m, 4m), instead of (0, 4m). Selena has learned a lot from this failure.
```
renepickhardt: Maybe we can add an infobox and express the information gain in bits? "learnt a lot" is a bit opinionated and even though we don't open a the math in this chapter I think it is worthwhile hinting that tools to express that stuff exist. We could refer to my privios paper for readers who wish to learn more

path_finding.asciidoc
```diff
@@ -223,6 +222,49 @@ Once we attempt to send an HTLC we can learn more about channel balances: if we
 You may be thinking "What's the point of learning from a successful HTLC?" After all, if it succeeded we're "done". But consider that we may be sending one part of a multi-part payment. We also may be sending other single-part payments within a short time. Anything we learn about liquidity is useful for the next attempt!
 ====
 
+==== Liquidity uncertainty and probability
+
+To quantify the uncertainty of a channel's liquidity, we can apply probability theory. A basic model of the probability of payment delivery will lead to some rather obvious, but important, conclusions:
+
+* Smaller payments have a better chance of successful delivery across a path.
```
renepickhardt: ...across a fixed path....

we don't loose anything by adding the adjective but it is actually more acurate. because depending on the path there might be paths where larger payments have a higher success chance the other paths with lower amounts.

path_finding.asciidoc
```diff
@@ -223,6 +222,49 @@ Once we attempt to send an HTLC we can learn more about channel balances: if we
...
+
+* Larger capacity channels will give us a better chance of payment delivery for a specific amount.
+
+* The more channels (hops), the lower the chance of success.
```
renepickhardt: maybe:  The more channels (hops) are included to deliver a payment, the lower the chance of success.

path_finding.asciidoc
```diff
@@ -223,6 +222,49 @@ Once we attempt to send an HTLC we can learn more about channel balances: if we
...
+
+More simply, if the possible values for the liquidity are 0,1,2,3,4,5 only one of those six possible values will be sufficient to send our payment. To continue this example, if our payment amount was 3, then we would succeed if the liquidity was 3, 4, or 5. So our chances of success are 3 in 6 (50%). Expressed in math, the success probability function for a single channel is:
+
+latexmath:[P_c(a) = (c + 1 - a) / (c + 1)]
+
+where +a+ is the amount and +c+ is the capacity
+
+From the equation we see that if the amount is close to 0, the probability is close to 1 whereas if the amount exceeds the capacity, the probability is zero.
+
+In other words: "Smaller payments have a better chance of successful delivery" or "Larger capacity channels give us better chances of delivery for a specific amount" and "You can't send a payment on a channel with insufficient capacity".
```
renepickhardt: it's a bit redundant with the three points at the beginning of the section but I think it is very fine to have that redundancy here.

path_finding.asciidoc
```diff
@@ -223,6 +222,49 @@ Once we attempt to send an HTLC we can learn more about channel balances: if we
...
+
+Now let's think about the probability of success across a path made of several channels. Let's say our first channel has 50% chance of success (P = 0.5). Then if our second channel has 50% chance of success (P = 0.5), it is intuitive that our overall chance is 25% (P = 0.25).
+
+We can express this as an equation that calculates the probability of a payment's success as the product of probabilities for each channel in the path(s):
+
+latexmath:[P_{payment} = \prod_{i=1}^n P_i]
+
+Where P_i_ is the probability of success over one path or channel, and P_payment_ is the overall probability of a successful payment over all the paths/channels.
```
renepickhardt: Better / more accurate: Where P_i_ is the probability of success over one channel, and P_payment_ is the overall probability of a successful payment over all channels in the (multi)path

path_finding.asciidoc
```diff
@@ -321,6 +351,10 @@ Therefore, Selena's node must consider how long to keep this knowledge before as
...
+
+One proposed solution to this problem is called "stuckless payments", and it depends on Point Time-Locked Contracts (PTLCs), which are payment contract that use a different cryptographic primitive than HTLCs (i.e. point addition on the elliptic curve instead of a hash and secret pre-image).
```
renepickhardt: I think it is important to note that 

1.) PTLCs basically require taproot. While there is the scriptless scripts way of doing this with ECDSA magic only with taproot and the resulting adaptor signatures we will go down this way 
2.) stuckless payments have there own issues / challanges the most important one is that you can decide to not settle an onion and thus cancle it. how ever the liquidity bound in that onion is still locked until the stuck payment returns. So in the worst case there will be a lot of bound liquidity

overall I am not sure if the note about stuckless payments should be moved to the HTLC / Routing chapter as this is not so much about path finding but more about transport of onions / HTLCs

renepickhardt (changes requested):
awesome text and discussion before. I have some tiny editorial improvements which I believe make it just more accurate but don't make it more complex for the reader

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+==== Trial-and-error over multiple "rounds"
+
+Multi-Path Payments lead to a somewhat modified "trial-and-error" loop for payment delivery. Because we are attempting multiple paths in each attempt, we have four possible outcomes:
+
+* All parts succeed, the payment is successful
+* Some parts succeed, some fail with errors returned
+* All parts fail with errors returned
+* Some parts are "stuck", no errors are returned
+
+In the second case, where some parts fail with errors returned and some parts succeed, we can now _repeat_ the trial-and-error loop, but *only for the residual amount*.
```
renepickhardt, line 411: and with the learnt knowledge from the uncertainty observations that we have discussed in the beginning of the payment delivery chapter

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+
+In our example we can see some immediate improvements to our path finding problem that become possible with MPP. First, we can utilize the S->X channel that has known insufficient liquidity to transport 1m satoshis plus fees. By sending a smaller part along that channel, we can use paths that were previously unavailable. Second, we have the unknown liquidity of the B->R channel, which is insufficient to transport the 1m amount, but might be sufficient to transport a smaller amount.
+
+===== Splitting payments
+
+The fundamental question is how to split the payments. More specifically, what is the optimal number of splits and the optimal amounts for each split?
+
+This is an area of ongoing research, where novel strategies are emerging. Multi-path payments lead to a different algorithmic approach than single path payments, even though single-path solutions can emerge from a multi-path optimization (i.e. a single-path may be the optimal solution suggested by a multi-path path finding algorithm).
```
renepickhardt, line 373: our research actually states how to do this optimally (depending of course on your optimization goal) selecting the proper optimization goal is and probably will be an active area of research but the general method / approach should not be switched. 
Also our mainnet test indicate that the optimization goal that we formulated is probably a rather desirebal one.

That being said I agree with the careful formulation here!

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+
+* The more channels you use, the chance of success becomes (exponentially) lower.
+
+From the first of these insights, we might conclude that splitting a large payment (e.g. 1 million satoshis) into tiny payments increases the chance that each of those smaller payments will succeed. The number of possible paths with sufficient liquidity will be greater if we send smaller amounts.
+
+To take this idea to an extreme, why not split the 1m satoshi payment into one million separate 1-satoshi parts? Well, the answer lies in our second insight: since we would be using more channels/paths to send our million single-satoshi HTLCs, our chance of success would drop exponentially.
+
+If it's not obvious, the two insights above create a "sweet spot" where we can maximize our chances of success: splitting into smaller payments but not too many splits!
+
+Quantifying this optimal balance of size/number-of-splits for a given channel graph is out of the scope of this book but it is an active area of research. Some current implementations use a very simple strategy of splitting the payment in two halves, four quarters etc.
```
renepickhardt, line 387: We could note from the example mentioned in my mailinglist post that this is obviously not optimal (c.f. the last paragraph of the mail: https://lists.linuxfoundation.org/pipermail/lightning-dev/2021-March/002984.html which I copy here for your convenience): 

```
One note on the question of MPP: In the last couple weeks I have been
collaborating with Christian Decker. I belief (by using the methodology
from this paper) to also have a definite solution to the question of:

How to split a payment into k parts and how many funds to allocate to each
path to increase the (multi)path success probability.

While this is is not addressed in the attached paper as we still need to
run evaluations I can already share that an equal sized split as used in
the paper (and by some implementations) is not preferable as one can easily
see from this example:
Imagine one is to deliver 100 satoshi and has to paths with 1 uncertain
channel on each path. The first of capacity 101 and  the second of 51.
Obviously trying to send 100 satoshi along the 101 capacity channel is bad.
Similarly splitting 50/50 and sending 50 Satoshi along the 51 satoshi
capacity channel is also bad. Thus a split that allocates for example 67
Satoshi to the 101 capacity and 33 to the 51 satoshi channel seems way more
reasonable. Actually 75/25 would probably be the best solution for such a
setting. And no it is only random coincident that a binary splitter would
have arrived at that split eventually (after potential miss trials)
There is way more math theory of how to actually solve the optimization
problem in the general case and how to find a split and paths that
maximizes the probability of the attempts. I cannot share these results yet
but I am pretty confident that there will be updates on that end very soon.
```

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+To read more about the optimization problem known as Minimum-Cost Flows involved when splitting payments into different sizes and allocating them to paths see the paper: https://arxiv.org/abs/2107.05322[Optimally Reliable & Cheap Payment Flows on the Lightning Network] by (co-author of this book) Rene Pickhardt & Stefan Richter.
+====
+
+In our example, Selena's node will attempt to split the 1m satoshi payment into two 500k satoshi parts and send them on two different paths. This is shown in <<mpp_paths>>:
+
+[[mpp_paths]]
+.Sending 2 x 500k parts of a multi-path payment
+image::images/mpp_paths.png[]
+
+Because the S->X channel can now be utilized, and (luckily for Selena), the B->R channel has sufficient liquidity for 500k satoshis, the two parts are successful along paths that were previously not possible.
```
renepickhardt: we might have adopted the example already to not split 1M into 500k / 500k as this transports the wrong inutition. we could do 750k along the S-A-B-R path and 250k along the S-X-Y-R path

renepickhardt: if wanted I could also just compute the optimal split fo such setting and provide you with those numbers. We don't have to explain how we arrived there but I think it makes sense to just present them for our setting

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+
+==== Trial-and-error over multiple "rounds"
```
renepickhardt, line 402: note (probably no need to change anything) this loop is actually very similar to the single path case only difference is we might get failure for a partial amount

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+* All parts fail with errors returned
+* Some parts are "stuck", no errors are returned
+
+In the second case, where some parts fail with errors returned and some parts succeed, we can now _repeat_ the trial-and-error loop, but *only for the residual amount*.
+
+Let's assume for example that Selena split her payment into 26 parts of varying sizes and of those 3 parts failed with errors. If those 3 parts consisted of, say 155k satoshis, then Selena would restart the path finding effort, only for 155k satoshis. The next "round" could find completely different paths (optimized for the residual amount of 155k), and split the 155k amount into complete different splits!
+
+[TIP]
+====
+While it seems like 26 split parts are a lot, tests on the Lightning Network have successfully delivered a payment of 0.3679 BTC by splitting it into 345 parts.
```
renepickhardt, line 419: and while 345 also seems a lot we note that optimal splitting minimizes the expected number of onions that one needs to send out in order to deliver the payment.

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+====
+
+Furthermore, Selena's node would update the channel graph using the information gleaned from the successes and errors of the first round, to find the most optimal paths and splits for the second round.
```
renepickhardt, line 422: ah see above I would move that information up! before I thought it was forgotten.

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+
+Let's say that Selena's node calculates that the best way to send the 155k residual is 6 parts split as 80k, 42k, 15k, 11k, 6.5k and 500 satoshis. In the next round, Selena gets only one error, indicating that the 11k satoshi part failed. Again, Selena updates the channel graph based on the information gleaned and runs the path finding again, to send the 11k residual. This time, she succeeds with 2 parts of 6k and 5k satoshis respectively.
```
renepickhardt, line 424: not sure if on our small example graph we would ever have such a split as bascially not that many paths exist. maybe best to stick to a split that is close to the example from the paper?

12_path_finding.asciidoc
```diff
@@ -0,0 +1,441 @@
...
+
+[[mpp_rounds]]
+.Sending a payment in multiple rounds with MPP
+image::images/mpp_rounds.pngp[]
+
+In the end, Selena's node used three rounds of path finding to send the 1m satoshis in 30 parts.
+
+=== Conclusion
+
+In this chapter we looked at path finding and payment delivery. We saw how to use the channel graph to find paths from a sender to a recipient. We also saw how the sender will attempt to deliver payments on a candidate path and repeat in a trial-and-error loop. Finally, we saw how the newly deployed Multi-Path Payments feature allows us to split payments into parts, increasing the probability of success even for larger payments.
```
renepickhardt: maybe emphasize once more that the uncertainty about the liquidity is part of the protocol for scalability reasonsn (c.f.: https://bitcoin.stackexchange.com/questions/107722/does-balance-broadcasting-hinder-scalability-of-ln/107723#107723 ) and for privacy reasons and that the payment delivery always consists of a trial and error loop with some probing. Overall delivering a payment is an optimization problem that can be solved depending on your optimization goal (lower fees, faster success, less nodes to be involved, ....)

renepickhardt (changes requested):
only looked at MPP chapter as I assume the other parts have not been changed while force pushing?

12_path_finding.asciidoc
aantonop: Agree, let's change this to an unequal split. I think having an equal split also undermines the learning objective

PR #734, opened 2021-07-08
Providing not proving

willtech3:

PR #751, opened 2021-07-21
fix typo chapter 05.

mknoszlig:


renepickhardt (commented):
has already been addressed by #734

PR #755, opened 2021-07-23
ch-07: construct revocation key derivation from first principles

Roasbeef:
In this commit, we attempt to construct the derivation method we used to
derive the revocation keys from first principles. This explanation was
inspired by a conversation with Rene, leading to a sort of Socratic
method explanation w.r.t why the remote party cannot be allowed to
unilaterally sign with the revocation public key.

This initial draft explanation is a bit too technical IMO, but it's
unclear how much prior knowledge we can expect from the reader and/or
how much space we should devote for explaining this little trick.

One thing we may want to do is use some shorter variable names (defined
earlier) to make this section a bit more compact.

### DIFF
--- a/07_payment_channels.asciidoc
+++ b/07_payment_channels.asciidoc
@@ -537,6 +537,25 @@ OP_CHECKSIG
 
 This is a conditional script (see <<conditional_scripts>>), which means the output can be spent if _either_ of the two conditions is met. The first clause allows the output to be spent by anyone who can sign for +<revocationpubkey>+. The second clause is timelocked by +<to_self_delay>+ blocks and can only be spent after that many blocks by anyone who can sign for +<local_delayedpubkey>+. In our example, we had set the +<to_self_delay>+ timelock to 432 blocks, but this is a configurable delay that is negotiated by the two channel partners. The +to_self_delay+ timelock duration is usually chose in proportion to the channel capacity, meaning that larger capacity channels (more funds), have longer +to_self_delay+ timelocks to protect the parties.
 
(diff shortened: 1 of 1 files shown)


07_payment_channels.asciidoc
```diff
@@ -533,7 +533,27 @@ OP_ENDIF
...
+
+To see why this works in practice, notice that we can _re order_ (commute) and expand the public key computation of the original formula for `revocation_basepoint`:
+```
+revocation_priv = G*(revocationbase_priv * sha256(revocation_basepoint || per_commitment_point) + G*(per_commitment_secret * sha256(per_commitment_point || revocation_basepoint))
+                 = G*(revocationbase_priv + sha256(revocation_basepoint || per_commitment_point) + per_commitment_point * sha256(per_commitment_point || revocation_basepoint))
+```
+
+In other words, the `revocationbase_priv` can only be derived (and used to sign for the `revocationpubkey` by the party that knows _both_ the `revocationbase_priv` _and_ the `per_commitment_secret`. This little trick si what makes the public-key based revocation system used in the Lightning Network secure.
```
renepickhardt: nit:  this little trick **IS** what makes ...

renepickhardt (changes requested):
we don't write w book about group theory and ECDSA stuff but I have the feeling that this tip could be simplified heavily if we go a bit more on meta and extend the theory. I could give it a try because I think in the current version it is already more accessible than the BOLTs but still probably very technical. 

I would do something like explaining that: 
1. pubkey = privkey * G
2. (pub1 + pub2) = (priv_1 + priv_2) * G
3. (G can be any point in the group (it does not have to be the generator)
4. pub1 + pub2 is the point / address (revocation pub key) to which the comitment tx commits and can be derived from sharing pubkeys in channel setup
5. noone know priv1 + priv2 unless someone share the private key which is exactly how what happens

Roasbeef:
Thanks for taking a look at this! 

> we don't write w book about group theory and ECDSA stuff but I have the feeling that this tip could be simplified heavily if we go a bit more on meta and extend the theory

So we have a section that explains how ECDH works in the section on onion routing, IIRC that section goes after this section, but possibly we could place it here instead? Given the same intuition is required to really grok how stuff like the ephemeral point blinding used in Sphinx works. 

Agree we don't want to go super into group theory, in my experience with developers new to this stuff, it's usually enough to show a long form expansion of the algebra, so they can see that everything reduces down to being able to do a public operation on that point, that can then be mirrored on the private data to derive the new key.

Roasbeef:
Rebased and fixed the typo.

PR #760, opened 2021-07-28
Fix typo 03_how_ln_works.asciidoc

paisalk:
Correct a typo from "In" to "If" on Line 11.


paisalk (commented):
Correct a typo from "In" to "If" on Line 11.

renepickhardt (approved):

PR #688, opened 2021-04-16
Update c-lightning docker version

kevin-miller-0405:
I got an error complaining the c-lightning version doesn't exist when trying out the example following chapter 4.
By updating to the latest version, I was able to get the example working.

PR #762, opened 2021-07-29
Docker: update c-lightning build version

dangershony:
The current defined version is not found 
https://launchpad.net/~lightningnetwork/+archive/ubuntu/ppa

### DIFF
--- a/code/docker/c-lightning/Dockerfile
+++ b/code/docker/c-lightning/Dockerfile
@@ -5,7 +5,7 @@ RUN apt update && apt install -yqq \
 	software-properties-common
 
 # c-lightning
-ENV C_LIGHTNING_VER 0.9.3~20210120202101201901~ubuntu20.04.1
+ENV C_LIGHTNING_VER 0.10.0~202104091415~ubuntu20.04.1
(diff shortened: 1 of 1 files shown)


csralvall:
Hello @dangershony, I was also thinking in a solution for this, but I wanted to avoid the dependence of the existence of a specific version in the PPA, because as we can see, all the older versions were removed from the PPA and we can't be sure that in the next upgrade of the package the current version will not follow the same luck.
I have implemented and tested my [idea](https://github.com/csralvall/lnbook/commit/2b8a86209f8c0092577d5ef5c0e2c33d2c2438c6) running the container inside a network with a bitcoin node and it works.
Also, I've checked that the version of the package inside the container was the last (v10 at the moment of writing this).
The explanation is in the commit message.

dangershony:
@csralvall I think your approach is probably better but the caveat is if there are breaking change in c-lightning (probably unlikely) it will break the docker workshop.

csralvall:
You are right. I will let the maintainer decide about this.

Also, as this issue was addressed with the same approach on older pull requests, I will tag them below (hope this helps).

Duplicate of #688

PR #778, opened 2021-08-10
Added extra cd command

Bosch-0:
Added extra cd command in the intro as to ensure people are inside the lnbook directory prior to executing the next commands later in the chapter.


renepickhardt (approved):
good catch

PR #786, opened 2021-08-14
Explain plugins

ghost:
The text says that "We have already explained how to activate plugins in c-lightning", but as far as I can see that's not true. I took a stab at it with a little note box.


ghost:
(I'm assuming you can squash commits when you merge, otherwise let me know and I'll remake the branch with just one patch)

renepickhardt (approved):
thanks!

PR #787, opened 2021-08-14
add ogunden contributor

ghost:
assuming you accept any of my other pulls..


renepickhardt (approved):
thanks for the autopilot addition and the lack of plugin description

PR #793, opened 2021-08-18
ch-10: add initial explanation of session key randomization

Roasbeef:
In this commit, we include an initial explanation of the session key
re-randomization scheme that makes the Sphinx packet format so compact.
Alice is able to generate all the necessary session keys up front by
using an initial session key, and doing an EC multiplication of that key
and a blinding factor derived from information of the last hop.

I wanted to use sub-scripts for a lot of the notation here, by the
asciidoc renderer I use didn't seem to be able to render them....

I think I want to make another pass and blend in the initial explanation
in the high level style of the initial ECDH explanation.


10_onion_routing.asciidoc
```diff
@@ -331,6 +331,26 @@ Thus, they will both get the same result which they can use as a shared key to s
 
 ****
 
+A unique trait of Sphinx as a mix-net packet format is that rather than include a distinct session key for each hop in the route, which would increase the size of the mix-net packet dramatically, instead a clever _blinding_ scheme is used deterministically randomize the session key at each hop.
```
Randy808: I think this sentence would read clearer this way (crossed out words represent removal and bolded words represent additions):
"A unique trait of Sphinx as a mix-net packet format is that rather than include a distinct session key for each hop in the route, which would increase the size of the mix-net packet dramatically, i̶n̶s̶t̶e̶a̶d̶ a clever _blinding_ scheme is used **to** deterministically randomize the session key at each hop."

10_onion_routing.asciidoc
```diff
@@ -331,6 +331,26 @@ Thus, they will both get the same result which they can use as a shared key to s
...
+The session key for hop `i` is derived using the node public key, and derived shared secret of hop `i-1`:
+```
+session_key_i = session_key_{i-1} * SHA256(node_pubkey_{i-1} || shared_secret_{i-1})
+```
+
+In other words, we take the session key of the prior hop, and multiply it by a value derived from the public key and derived shared secret for that hop.
+
+As EC multiplication can be performed on a public key without knowledge of the private key, each hop is able to re-randomize the session key for the next hop in a deterministic fashion.
+
+The creator of the onion packet knows all the shared secrets (as they're encrypted the packet uniquely for each hop), and thus is able to derive all the blinding factors.
```
Randy808: Edit for clarity (plural verb form for singular 'they'):
"The creator of the onion packet knows all the shared secrets (as they've encrypted the packet uniquely for each hop), and thus i̶s̶ **are** able to derive all the blinding factors."

Reference: https://apastyle.apa.org/style-grammar-guidelines/grammar/singular-they

10_onion_routing.asciidoc
```diff
@@ -642,14 +673,19 @@ No one can tell the difference between filler put there by Alice and filler put
...
+```
 
-Clarify how the HMAC and session key are computed for this onion packet
+With this generated, Bob now re-randomizes the session key by performing an EC multiplication using his session key and the blinding factor:
+```
+session_key_chan = session_key_bob * b_bob
+```
 
-////
+The `session_key_chan` public key will then be appended to the front of the onion packet for processing by Chan>
```
Randy808: There's a typo at the end of this line, I think you meant '.' instead of '>'

10_onion_routing.asciidoc
Roasbeef: fixed!

10_onion_routing.asciidoc
Roasbeef: Thanks, accepted the suggestion

10_onion_routing.asciidoc
Roasbeef: Yep, slight typo

Roasbeef:
Comments addressed! Just pushed up a fresh version, landing this now so it makes some of the latest more sweeping edits easier merge conflict wise.

PR #796, opened 2021-08-18
Add HMAC to glossary

juhi24:
Added a glossary entry for HMAC since the abbreviation is used in another entry.


renepickhardt (changes requested):
It is actually not only used in onion routing but also in the noice protocol for lightning messages

Roasbeef:
Committed as `b0128c4c6de224cb34f46357eb83d0ffcc43d4d8` (with Rene's edit)

PR #825, opened 2021-08-19
fix typo and correct text formatting

ghost:


renepickhardt (changes requested):
I actually think that it would be better if we write: 

In our case the weights of the edges might represent the routing fees.

ghost:
Your suggestion sounds fine to me. Note that the "typo fix" I made here (sorry for the big diff, that's due to the formatting change) is to remove an extra "or" and add a period "." after A-star.

PR #830, opened 2021-08-19
you probably want to settle on 'LN' rather than 'the LN'

ghost:
There's an inconsistency in this chapter, sometimes the network is referred to as "the LN" and other times just as "LN". This PR makes it always just "LN".

This is one option, the other is coming in the next PR.


renepickhardt:
I leave this open as I belive this is global copy editing ?

PR #831, opened 2021-08-19
alternatively, maybe you want to settle on 'the LN'

ghost:
see discussion in #830 . you only want to merge this one or that one, not both.


renepickhardt:
note there is also #830  for our convenience!

@ogunden : great service! thanks

aantonop:
We might end up writing it out in full. I'll leave this to the O'Reilly copy editor and proof reader

PR #837, opened 2021-08-25
Update the c-lightning installation instructions and make things a bit more stable

cdecker:
This brings c-lightning up to version 0.10.1, using the reproducibly built binaries and verifying them. Funding c-lightning was broken due to the default wallet not being present in `bitcoind` anymore, and thus the JSON-RPC call failing.

~Still open:~
 - [ ] ~Payment fails because the channel between `Carol` and `Dina` being announced only in one direction~


aantonop:
Thanks for the PR @cdecker 

I've been working on some updates to the docker files on a separate branch, using the docker hub elements repository, which you maintain:

```
FROM elementsproject/lightningd:${CL_VER}
```

I noticed that you are not using that, instead pulling the binary from github releases. I assume that's because the docker container might not be updated as often? 

Should we prefer the github release files and build our own docker or use the docker image you maintain?

cdecker:
The reproducible builds should generally be preferred because we can ensure that the binaries actually match the source code, and we can attest that the binaries were built correctly. The PPA requires us to build in a dynamic environment on launchpad's machines, resulting in slight differences in the binaries that prevent us from creating those attestations. I should really switch the `elementsproject/lightning` docker images to use those instead, however we're slowly weening off of docker hub, since they started severely limiting access to popular images.

aantonop (approved):

PR #841, opened 2021-09-02
fix: typo exmaple -> example

woodyjon:
Fixed a typo exmaple -> example.

Also moved the name of a contributor to right place in alphabetical order.

I hope it's not too late for the production.
And if it's too small to be added in the contributors' list, tell me, I'll remove it.


Roasbeef:
Fixed in the latest version, thanks for the PR in any case!

PR #861, opened 2021-09-20
fix typos

ghost:


aantonop:
Current status is "IN PRODUCTION". We are not accepting PRs until the book is through copy-edit, about 2 months.

PR #862, opened 2021-09-21
RaspiBlitz service list updated

CandleHater:
Updated the service list of the RaspiBlitz to the latest state.


renepickhardt (changes requested):
maybe we could make this list in alphabetical order. there seem to be some parts from the old list to be removed. How was the list chosen? just copied from raspi blitz documentation or is it opinionated?

aantonop:
This is unnecessary detail for this section. We are not advertising these as products. A short list of the most important capabilities serves to illustrate the range of services. It doesn't need to be complete.

PR #865, opened 2021-09-21
lighting corrections

RandyMcMillan:


RandyMcMillan:
I understand that the current version is in production but the use of 'lighting' seems to be a critical correction.

Roasbeef (approved):
heh even after all these years I _still_ find it easy to make this mistake 

LGTM 🏘

PR #870, opened 2021-09-22
Update glossary entry for address

murchandamus:
* Uses a bech32 address in the example
* Does not compare addresses with email addresses which suggests that addresses should be reused


Roasbeef (approved):
Thanks for all these PRs! 

LGTM ☔️

PR #871, opened 2021-09-22
Update bech32 glossary entry

murchandamus:
* Properly distinguish between native segwit outputs and bech32 addresses
* Drop mention of the 90 character limit that is ignored by lightning invoices
* Mention that Lightning invoices use bech32
* Point out that higher version native segwit outputs use bech32m


Roasbeef (approved):
LGTM 🛒

PR #872, opened 2021-09-22
Update blockchain, coinbase, and coinbase tx

murchandamus:
* Previous phrasing sounds as if blocks could only be found every ten minutes which may or may not happen
* The coinbase input does not collect the mining reward, it is claimed by assigning the reward to one or more outputs in the coinbase transaction


Roasbeef (approved):
LGTM 💣

PR #873, opened 2021-09-22
Update glossary entry for block

murchandamus:
Previous phrasing could be understood that proof of work makes transactions valid, but rather a block may only include valid transactions and confirms them.


Roasbeef (approved):
LGTM 📌

PR #880, opened 2021-09-29
images: Remove artifact from anon-set

maflcko:
Remove "Window Snip", which seems an oversight


maflcko:
Also, used zopflipng for compression

maflcko:
Why the close?

aantonop:
Book is in production since last month and OReilly is redoing all
illustrations and copy editing.

We will merge all their changes in a few weeks, so we don't want to
conflict with changes here

PR #884, opened 2021-10-02
Update 07_payment_channels.asciidoc

timohanke:
Something in the equations for revocationpubkey and revocation_priv was mixed up.


Roasbeef (approved):
Nice catch! The changes here a correct, also the prior equation was actually self referential (`per_commitment_secret` was defined in terms of `per_commitment_secret`) 🤦🏾‍♂️

PR #930, opened 2022-01-16
Typo: 24 instead of 64

stickies-v:
As per a few sentences earlier: "(24 bits of the locktime and 24 bits of the sequence)"


morehouse:
See also https://github.com/lnbook/lnbook/pull/960 to correct a related inaccuracy in encoding scheme.

PR #931, opened 2022-01-16
Remove duplicated explanation of distinct session keys

stickies-v:
This paragraph is repeated from a (more detailed) explanation just a few paragraphs up.


Donaldson7 (approved):
Want to get full tools to work on projects

PR #945, opened 2022-02-19
Update to LND Dockerfile

IceXPR:
Dockerfile didn't build image using the checked in version of the code.  

These are PR code changes to fix it:
-GO to 1.17
- LND to v0.14.2-beta
- uses Docker WORKDIR to run make instead of changing directories.
- pulls the code from GitHub using the tag and store in the proper directory in one step
- uses git clone with a depth of 1 as the GitHub history is not needed when building

### DIFF
--- a/code/docker/lnd/Dockerfile
+++ b/code/docker/lnd/Dockerfile
@@ -1,6 +1,6 @@
 ARG OS=ubuntu
 ARG OS_VER=focal
-ARG GO_VER=1.13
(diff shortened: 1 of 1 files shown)


NicolaLS:
This fixed `docker build -t lnbook/lnd lnd` which failed previously for me. However running `bash run-payment-demo.sh` now hangs but that was fixed by #970 after that the demo succeeds!

PR #958, opened 2022-05-07
update link 'ln.bigsun.xyz' => 'ln.fiatjaf.com'

DeeterCesler:
bigsun.xyz is abandoned and fiatjaf has moved to a new domain, [as his GitHub confirms](https://github.com/fiatjaf).


vv01f:
s. #922

PR #970, opened 2022-07-02
Add dependency of lightning nodes to bitcoind

alexandre-k:
When starting all docker containers with `docker compose up`, there are errors from lightning nodes happening, particularly Alice where the start of the LND node results in an error of type `panic: Unable to synchronize wallet to chain`.
A solution is to make the lightning nodes depends on bitcoind to start up.


NicolaLS (approved):
LGTM

alexandre-k:
Thanks.

